PR #8, opened 2021-09-29
Accessibility improvements

stevenzeck:
This PR aims to fix a few accessibility issues.

1. In the OPDS feeds, replaced the button text "More..." with a forward arrow to be more standard with other Android apps.  It doesn't go all of the way to the right like it should, but that entire screen still needs to be redone with a proper XML layout, unless we adopt Jetpack Compose sooner than later.
2. In the highlight context menu, replaced the existing edit and delete button icons with filled ones to be more consistent with the rest of the test app.
3. Also in the highlight context menu, made all touch targets 48dp x 48dp, as that is the minimum recommended width x height.  I haven't really used `layer-list` and `shape`, so I imagine those selector files can still be optimized better.
4. In the audiobook player, made the 30 second backward and forward buttons 48dp x 48dp.
5. Added `android:contentDescription` attributes that were missing across the entire app


mickael-menu (approved):
Thanks, and congrats on the first real PR of this repo! 😄

PR #13, opened 2021-10-01
Fix EpubNavigatorFragment.goBackward/Forward()

mickael-menu:
#### Fixed

* `EpubNavigatorFragment`'s `goForward()` and `goBackward()` are now jumping to the previous or next pages instead of resources.

The implementation was not following the spec of `Navigator`, since it was jumping between resources instead of pages.

```kotlin
    /**
     * Moves to the next content portion (eg. page) in the reading progression direction.
     */
    fun goForward(animated: Boolean = false, completion: () -> Unit = {}): Boolean

    /**
     * Moves to the previous content portion (eg. page) in the reading progression direction.
     */
    fun goBackward(animated: Boolean = false, completion: () -> Unit = {}): Boolean
```

I also made the conformance to `R2BasicWebView.Listener` private in `EpubNavigatorFragment`, to avoid future conflicts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,14 @@ All notable changes to this project will be documented in this file. Take a look
 
 **Warning:** Features marked as *alpha* may change or be removed in a future release without notice. Use with caution.
 
-<!--## [Unreleased]-->
+## [Unreleased]
+
+### Navigator
+
+#### Fixed
+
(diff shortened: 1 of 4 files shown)


qnga (approved):
Sounds good! I was not aware of the intended behaviour of `goBack` and `goForward`. It makes more sense this way.

mickael-menu:
Me neither, I'm not sure this was really used since the edge tap to turn pages is currently handled in the navigator. But I tried implementing page turns using the volume buttons and noticed this problem.

PR #14, opened 2021-10-04
Fix turning pages with an odd count of columns

mickael-menu:
#### Navigator

* Fixed turning pages of an EPUB reflowable resource with an odd number of columns. A virtual blank trailing column is appended to the resource when displayed as two columns.

For an EPUB reflowable resource, having an odd number of columns when displaying two columns per screen causes snapping and page turning issues. To fix this, we insert a blank virtual column at the end of the resource.


aferditamuriqi:
> @aferditamuriqi I'm going to merge this PR by the end of the week unless you report any problem with this. We had this fix in production for about two weeks now. If an issue pops up, we can always revert these changes later on.

Hey @mickael-menu I will be trying to test this before the end of the week, if i can't test it within a current implementation this week, for sure next week. i really would like to see if the last reading position stays accurate between scroll mode and paginated mode when adding an additional column.

aferditamuriqi:
@mickael-menu i did test this in terms of last reading position and it does falsify it, but i think it's a bit more complicated than that. we can maybe discuss this in more detail tomorrow on the call or when you have some times.

aferditamuriqi:
@mickael-menu there was actually a bigger issue with odd numbers on two column screens, the tap to go to the next page  was stuck when there was an odd number. i do think that this PR fixes majority of the issues around page numbers, snapping and swiping. the only thing i am concerned about that it falsifies the progression. maybe we need to change how the progression is calculated and base it on the column numbers instead of screen numbers. that could i think fix all the issues. but more to discuss i think :)

aferditamuriqi (approved):

mickael-menu:
@aferditamuriqi Yeah we can raise the subject during next call. I think it would be interesting to save the text context as well, now that we have a way to jump to a given text. This would be less fragile than the progression and might work better if the publication content changes.

aferditamuriqi:
> @aferditamuriqi Yeah we can raise the subject during next call. I think it would be interesting to save the text context as well, now that we have a way to jump to a given text. This would be less fragile than the progression and might work better if the publication content changes.

I agree.

PR #18, opened 2021-10-07
Add support for horizontal reading progressions for PDF

mickael-menu:
#### Added

* The PDF navigator now honors the publication reading progression with support for right-to-left and horizontal scrolling.
    * The default (auto) reading progression for PDF is top-to-bottom, which is vertical scrolling.

With the upcoming presentation settings, we'll be able to change the reading progression from the user settings.


qnga:
I don't understand why the default reading progression should be TTB... In my mind, scrolling mode and reading progression are two (mostly) different things.

qnga:
EPUBs can be read continuously too. With your interpretation, I feel like the `readingOrder` property of Epub-originated publications should not necessarily be the same as the `spine`'s `page-progression-direction`. We might have `TTB` or `BTT`, at will. I'm afraid all of this might make little sense.

PR #23, opened 2021-10-15
Fix overriding viewport meta tags in reflowable resources

### LINKED ISSUE #21: EPub is rendered different in Readium 2 than with Readium 1
## Bug Report

We have an epub that was rendered fine with Readium 1, but after upgrading to Readium 2 it renders really strange.

### What happened?

Rendered in Readium 1:
image

Rendered i Readium 2:
image

The publication has been shared through private channel

mickael-menu:
Fix #21 

The meta tag is now injected through JavaScript instead of directly modifying the resource.


qnga (approved):
Looks good.

PR #28, opened 2021-10-19
Add callback for handling internal links

mihai-wolfpack:


stevenzeck:
I think this is fine for now.  I wanted to see what happens if it can't find the file that matches the URL of the internal link.  In that case, it logs an error and the reader simply goes to the first resource of the book.

mickael-menu:
Ha, probably something to handle in `go()` to abort (and return `false`) if the `link.href` can't be found in the `readingOrder`. It's not directly related to this PR though, so not a blocking issue for the review. I guess the same would happen if a table of contents has a broken link.

PR #42, opened 2021-11-04
Enable source and doc publishing

qnga:
Allow to publish sources and documentation to Maven. I checked this can be used  by hand, and I _think_ it will be usable with Jitpack as well. It would make debugging much easier in apps using Jitpack.


mickael-menu (approved):
Looks useful, thanks!

PR #43, opened 2021-11-05
Add a fallback when Chrome Custom Tabs are not available

mickael-menu:
Opening an external link in an EPUB failed if Chrome Custom Tabs were disabled on the device.


readium/shared/src/main/java/org/readium/r2/shared/OptIn.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package org.readium.r2.shared
+
+@RequiresOptIn(
+    level = RequiresOptIn.Level.ERROR,
+    message = "This is an internal API that should not be used outside of Readium modules. No compatibility guarantees are provided."
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.TYPEALIAS, AnnotationTarget.PROPERTY)
+annotation class InternalReadiumApi
```
mickael-menu, line 18: Got inspired by [`InternalCoroutinesApi`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-internal-coroutines-api/).

PR #44, opened 2021-11-05
Fix OutOfMemoryError during media type sniffing

mickael-menu:
We got some reports of `OutOfMemoryError` crashes while sniffing a file as a string:

Screen Shot 2021-11-05 at 12 07 31


qnga (commented):
Looks good! Is https://github.com/readium/kotlin-toolkit/blob/966d1d07a88c3fcbfe5b0f3f747fcc1a94872862/readium/shared/src/main/java/org/readium/r2/shared/util/mediatype/SnifferContent.kt#L38 still useful though?

mickael-menu:
@qnga Good question, I think it can't hurt to prevent going into a `OutOfMemoryError` in the first place since when we use this, it's always to read a text file. It should improve performance too.

qnga (approved):

PR #55, opened 2021-11-08
Add `EdgeTapNavigation` to trigger page turns on edge taps

mickael-menu:
#### Navigator

* A new convenience utility `EdgeTapNavigation` to trigger page turns while tapping the screen edges.
    * It takes into account the navigator reading progression to move into the right direction.
    * Call it from the `VisualNavigator.Listener.onTap()` callback as demonstrated below:
    ```kotlin
    override fun onTap(point: PointF): Boolean {
        val navigated = edgeTapNavigation.onTap(point, requireView())
        if (!navigated) {
            // Fallback action, for example toggling the app bar.
        }
        return true
    }
    ```

---

This utility is useful for two reasons:

* We want to let the apps have total control over the tap events, and decide whether or not to handle page turns with edge taps.
* Implementing the edge taps properly (taking into account the reading progression) is not so easy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,18 @@ All notable changes to this project will be documented in this file. Take a look
 
 * The PDF navigator now honors the publication reading progression with support for right-to-left and horizontal scrolling.
     * The default (auto) reading progression for PDF is top-to-bottom, which is vertical scrolling.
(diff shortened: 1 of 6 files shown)


qnga (approved):
Makes sense!

PR #58, opened 2021-11-18
Fix building with Kotlin 1.6

mickael-menu:
### Fixed

* Fix building with Kotlin 1.6.

---

This doesn't bump the minimum required Kotlin version to 1.6, but fixes any build issue if someone chooses to use it.

### DIFF
--- a/readium/lcp/src/main/java/org/readium/r2/lcp/license/LicenseValidation.kt
+++ b/readium/lcp/src/main/java/org/readium/r2/lcp/license/LicenseValidation.kt
@@ -280,6 +280,7 @@ internal class LicenseValidation(
                     is State.registerDevice -> registerDevice(state.documents.license, state.link)
                     is State.valid -> notifyObservers(state.documents, null)
                     is State.failure -> notifyObservers(null, state.error)
+                    State.cancelled -> notifyObservers(null, null)
                 }
             }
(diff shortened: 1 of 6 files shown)


readium/lcp/src/main/java/org/readium/r2/lcp/license/LicenseValidation.kt
```diff
@@ -315,6 +316,8 @@ internal class LicenseValidation(
     private suspend fun fetchStatus(license: LicenseDocument) {
         val url = license.url(LicenseDocument.Rel.status, preferredType = MediaType.LCP_STATUS_DOCUMENT).toString()
         // Short timeout to avoid blocking the License, since the LSD is optional.
+        // FIXME: To remove after bumping to Kotlin 1.6
+        @Suppress("DEPRECATION")
         val data = network.fetch(url, timeout = Duration.seconds(5))
```
mickael-menu, line 321: I couldn't find any API to create a `Duration` in seconds that doesn't give a deprecation warning either on Kotlin 1.5 or 1.6... Since we have the `android.kotlinOptions.allWarningsAsErrors` flag on in `build.gradle`, we need to suppress the deprecation warning for now to be able to build.

qnga (approved):

PR #60, opened 2021-11-24
Add API `Navigator.Listener.onJumpToLocator()`

mickael-menu:
#### Navigator

* The new `Navigator.Listener.onJumpToLocator()` API is called every time the navigator jumps to an explicit location, which might break the linear reading progression.
    * For example, it is called when clicking on internal links or programmatically calling `Navigator.go()`, but not when turning pages.
    * You can use this callback to implement a navigation history by differentiating between continuous and discontinuous moves.

---

Note: This is a quickfix for the shortcomings of the current Navigator API which can't be used to build a navigation history. In the [future navigator](https://github.com/readium/kotlin-toolkit/discussions/51) I think we should emit more information about a navigation event, such as the source (link clicked, `go()` call, page turned, skip forward, etc.).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ All notable changes to this project will be documented in this file. Take a look
         return true
     }
     ```
-* The new `Navigator.Listener.onJumpToLocator()` API is called every time the navigator will jump to a discontinuous location.
+* The new `Navigator.Listener.onJumpToLocator()` API is called every time the navigator jumps to an explicit location, which might break the linear reading progression.
     * For example, it is called when clicking on internal links or programmatically calling `Navigator.go()`, but not when turning pages.
     * You can use this callback to implement a navigation history by differentiating between continuous and discontinuous moves.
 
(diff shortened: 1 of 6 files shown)


qnga (approved):
Maybe you should be more explicit about the callback: it is called if the move is intended to be discontinuous, though it could turn out to be continuous or no move.

mickael-menu:
> Maybe you should be more explicit about the callback: it is called if the move is intended to be discontinuous, though it could turn out to be continuous or no move.

It doesn't really matter, the idea is that the user triggered an action that breaks the continuous flow of progression, but it could end up in the same page or go to the next one.

Do you have any alternative for the API name that could better represent that?

qnga:
Oh no, I was thinking only about documentation, since I wondered about that and had to look at the code to get the answer.

mickael-menu:
I see, I will try to clarify it tomorrow 👍

mickael-menu:
@qnga Is it clearer?

qnga:
Yes!

PR #61, opened 2021-11-26
Support for checking a Publication Profile conformance

mickael-menu:
#### Shared

* A new `Publication.conformsTo()` API to identify the profile of a publication.
* Support for the [`conformsTo` RWPM metadata](https://github.com/readium/webpub-manifest/issues/65), to identify the profile of a `Publication`.

#### Shared

* `Publication.type` is now deprecated in favor of the new `Publication.conformsTo()` API which is more accurate.
    * For example, replace `publication.type == Publication.TYPE.EPUB` with `publication.conformsTo(Publication.Profile.EPUB)` before opening a publication with the `EpubNavigatorFragment`.


readium/shared/src/main/java/org/readium/r2/shared/publication/Metadata.kt
```diff
@@ -65,6 +66,7 @@ data class Metadata(
     constructor(
         identifier: String? = null, // URI
         type: String? = null, // URI (@type)
+        conformsTo: List<Publication.Profile> = emptyList(),
```
qnga, line 69: I would use a `Set` since the order is not meaningful.

mickael-menu: Good point, let me fix that.

qnga (approved):

PR #64, opened 2021-12-03
Fix the rendering of PDF covers in some edge cases

mickael-menu:
#### Streamer

* Fixed the rendering of PDF covers in some edge cases.


readium/streamer/src/main/java/org/readium/r2/streamer/parser/pdf/PdfiumDocument.kt
```diff
@@ -54,22 +54,20 @@ internal class PdfiumDocument(
 }
 
 private fun PdfiumCore.renderCover(document: _PdfiumDocument): Bitmap? {
-    try {
-        val pointer = openPage(document, 0)
-        if (pointer <= 0) return null
```
mickael-menu, line 59: I'm not sure why I checked for the pointer to be above 0, but sometimes it is negative and still valid. This prevented a few PDF covers to be rendered on some devices.

qnga: 0 is supposed to be an invalid pointer. I'm not sure the method can return it here though.

mickael-menu: Yes that's weird. Maybe the `openPage()` fails for some reason, but that doesn't prevent the rest of the code from working as we don't use the `pointer` directly.

qnga: I suppose in case of failure you won't get the first page though...

mickael-menu: In practice yes, but that's not guaranteed... I will look again on Friday to see if I can find a better way.

mickael-menu: @qnga I investigated more and looking at the Java and C++ code of PdfiumAndroid, I don't think the returned long is the address of a C++ pointer. I also tried giving a non-existing page index and this function throws an exception, so the error handling is properly handled inside.

So I think this PR is safe to merge as is. We also used this code in production for the past two weeks without bug reports.

qnga (approved):

PR #67, opened 2021-12-10
Introduce `adapters` modules for heavy dependencies

mickael-menu:
This PR extracts the code using heavy third-party dependencies into their own Gradle modules, under `readium-adapter-*`.

### Why?

Even though Readium aims to be modular, each component is currently still quite monolithic. For example, using `readium-navigator` pulls all the dependencies of each media type (AndroidPdfViewer, ExoPlayer for audiobooks, etc.). It's an issue because:

* Some dependencies increase significantly the size of the app bundle even if you don't want to support PDF, audiobook, CBZ, etc.
* If one of the third-party libraries Readium depends on breaks, the whole toolkit breaks, instead of just the module that needs this dependency.
* If an app wants to switch a navigator implementation using a different library (e.g. PSPDFKit), they still get AndroidPdfViewer pulled in their dependencies.

### What are the new modules?

* `readium-adapter-nanohttpd`
    * Our current HTTP server implementation using NanoHTTPD.
    * To ensure backward-compatibility, this is implicitly imported by `readium-streamer`, until the next major version.
* `readium-adapter-androidx-media`
    * Generic implementation of an Audio Navigator using Jetpack `media`.
    * Apps using the experimental audio navigator will need to add this new dependency to continue using it.
* `readium-adapter-exoplayer`
    * Concrete implementation of the `readium-adapter-androidx-media`'s `MediaPlayer` interface using ExoPlayer.
    * Apps using the experimental audio navigator will need to add this new dependency to continue using it.
* `readium-adapter-pdfium`
    * Imports `readium-adapter-pdfium-document` and `readium-adapter-pdfium-navigator`.
    * Apps using the experimental PDF navigator will need to add this new dependency to continue using it.
* `readium-adapter-pdfium-document`
    * Implementation of a PDF parser using `PdfiumAndroid`.
* `readium-adapter-pdfium-navigator`
    * Implementation of a PDF navigator using `AndroidPdfViewer`.

### What's next?

This PR focuses on the heavy dependencies, but I think we should go further with this approach and separate clearly the core interfaces and models from the concrete implementations for each media type. For example:

* `readium-core-parser` – Interfaces and utilities to write Readium parsers
* `readium-core-navigator` – Interfaces and utilities to write Readium navigators
* `readium-epub` – Imports `readium-epub-*`
    * `readium-epub-parser` – Parser for EPUB publications
    * `readium-epub-navigator` – Navigator for EPUB publications


mickael-menu:
I'm postponing this PR for now, things to do:

* Move the `PdfNavigatorFragment` back to `readium-navigator`, and add `PdfView` + `PdfViewFactory` interfaces.
* Implement `PdfView` in `readium-adapter-pdfium-view`.

mickael-menu:
Superseded by https://github.com/readium/kotlin-toolkit/pull/129

PR #69, opened 2021-12-14
Update dependencies and fix tests

stevenzeck:
I started out just updating dependencies, but decided to fix some of the issues with unit tests and json.

- Updated Kotlin from 1.5.31 to 1.6.10
- Updated Gradle from 7.0.3 to 7.0.4
- Updated all dependencies in all modules.  This required modifying some of the code to fix breaking changes, but should not affect functionality
- Using the `org.json:json` library for mocking JSONObject and the like in the tests was not working on newer versions.  As it is already using `org.robolectric:robolectric` I annotated each relevant file with `@RunWith(RobolectricTestRunner::class)`
- Removed the `jackson` dependencies and deleted the `MediaOverlayHandler` class
- Added an index in the Highlight table for `BOOK_ID`, as that warning has been there for several months
- Cleaned up test dependencies, removing unused ones and migrating some assertion calls from one framework to another


test-app/src/main/java/org/readium/r2/testapp/domain/model/Highlight.kt
```diff
@@ -30,6 +30,7 @@ import org.readium.r2.shared.util.MapWithDefaultCompanion
     foreignKeys = [
         ForeignKey(entity = Book::class, parentColumns = [Book.ID], childColumns = [Highlight.BOOK_ID], onDelete = ForeignKey.CASCADE)
     ],
+    indices = [Index(value = [Highlight.BOOK_ID])]
```
mickael-menu, line 33: Nicely spotted ;)

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt
```diff
@@ -139,6 +139,9 @@ open class ReaderActivity : AppCompatActivity() {
             is ReaderViewModel.Event.Failure -> {
                 Toast.makeText(this, event.error.getUserMessage(this), Toast.LENGTH_LONG).show()
             }
+            else -> {
+                Toast.makeText(this, getString(R.string.error), Toast.LENGTH_LONG).show()
+            }
```
mickael-menu, line 144: This will show the toast when the user start a search in a book. I think instead we should simply add branches for all the available events.

Currently the `SearchFragment` is handling the `StartNewSearch` event which is dangerous as if we revert to a single `Channel` for event handling, it can by default have only a single observer. Best practice would be to handle all the events in the `ReaderActivity`, and forward downward any event if needed.

For this PR, I suggest we just move the `StartNewSearch` event to the `ReaderViewModel.FeedbackEvent` enum and let the `EpubReaderFragment` take ownership of this event handling, as it has access to the `SearchFragment`.

test-app/src/main/res/values/strings.xml
```diff
@@ -197,4 +197,6 @@
     <string name="play_or_pause">Play or pause</string>
     <string name="go_forward_30_seconds">Go forward 30 seconds</string>
     <string name="select_a_font">Select a font</string>
+
+    <string name="error">Error</string>
```
mickael-menu, line 201: To be removed.

mickael-menu (changes requested):

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt
stevenzeck: `EpubReaderFragment` doesn't have a reference to `SearchFragment`, all it does show the fragment via the fragment manager, so it's not quite as easy as letting EpubReaderFragment take ownership of event handling for this.  But yeah that's an overlook on my part putting a toast error message when it results in showing that when a new search begins.  I think the best thing to do is just add an empty else branch.

test-app/src/main/res/values/strings.xml
stevenzeck: Yeeted

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt
mickael-menu: Sounds good for now, we can always fix this in a dedicated refactoring.

mickael-menu (approved):

PR #71, opened 2021-12-17
Replace the local HTTP server

dotmonkey:
http server is not needed now for EPUB publications.

### DIFF
--- a/readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2EpubPageFragment.kt
+++ b/readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2EpubPageFragment.kt
@@ -13,6 +13,7 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.content.SharedPreferences
 import android.graphics.PointF
+import android.net.Uri
 import android.os.Bundle
 import android.util.Base64
 import android.util.DisplayMetrics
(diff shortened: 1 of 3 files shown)


mickael-menu (changes requested):
Thanks for taking a look at this issue, that's a nice first shot.

There's a number of issues to address before we can merge this feature in:

* The `Publication` should not expose a new public `serverFetcher` property.
* The code handling the injection currently in the server should be moved to the navigator (`HtmlInjector`) instead of relying on the current `Server` implementation, which will be deprecated.
* Byte range requests don't seem to be handled in the HTTP response, which will be important for serving video or audio with EPUB 3.
* To serve the assets, maybe `WebViewAssetLoader` would be more suitable?

dotmonkey:
Thanks for review.
some reasons about the issues: 
- HtmlInjector is now in streamer module, and not visible to navigator module. 
- HtmlInjector needs field userPropertiesPath and customResources which is also not available in class R2EpubPageFragment or EpubNavigatorFragment.
- when range filed of ResourceInputStream object is null,ResourceInputStream will auto detect the length for the resource. wether browser read a byte or ByteArray one time is determined by the browser.
- WebViewAssetLoader just process one path:assets, and it use URLConnection.guessContentTypeFromName to guess mime type, so js and font files will got null mime type.

mickael-menu:
Thanks for the additional details,

I'm working on a new Presentation API which will render `userPropertiesPath` obsolete. It will probably be merged early next year, I think we can revisit your PR then to use this new API instead which will remove any dependency on the streamer.

dotmonkey:
OK

PR #74, opened 2021-12-20
Migrate to Kotlin DSL for Gradle

stevenzeck:
This PR migrates the Gradle files from Groovy to Kotlin DSL.


stevenzeck:
@mickael-menu Feel free to do a pre-review while in Draft state.

stevenzeck:
This is as good as it's going to get for now.  Per discussion in Slack, eventually we want to use a centralized versioning system so that any modules that depend on the same libraries all reference the same version.  See https://docs.gradle.org/current/userguide/platforms.html#sub:central-declaration-of-dependencies.

The wild card is Dokka.  It runs locally, so we'll have to see what it does when it goes through CI.

mickael-menu (approved):
Looking good, thanks!

PR #80, opened 2022-01-03
Implement a new audiobook navigator based on Jetpack Media2

qnga:
#### Navigator

* (*experimental*) A new audiobook navigator based on Jetpack `media2`.
    * See the [pull request #80](https://github.com/readium/kotlin-toolkit/pull/80) for the differences with the previous audiobook navigator.
    * This navigator is located in its own module `readium-navigator-media2`. You will need to add it to your dependencies to use it.
    * The Test App demonstrates how to use the new audiobook navigator, see `MediaService` and `AudioReaderFragment`.
    
---
This new audiobook navigator departs from the previous one in the following way:
- No `MediaController` is used.
- No `Service` provided and no UI-related stuff (including for the notification zone).
- Apps must build a `MediaControllerSession`, use a helper to get a session from it, and connect that session to their media service.

Secondarily, this PR:
- Moves the logic about saving the last location from `BaseFragment` to `VisualFragment` because the media service is responsible for this in the audiobook case. The stuff about decorators is moved as well to `VisualFragment` because it involves visual responses to the decorator API.
- Reverts some changes from commit 553c7d3e4013 which has broken the separation between the visual and non-visual cases. Fixes the padding issue in search fragment in a different way.
- Implements a `ReaderRepository` and gets rid of intent publication helpers.


test-app/src/main/java/org/readium/r2/testapp/reader/AudioReaderFragment.kt
```diff
@@ -1,53 +1,53 @@
...
 import org.readium.r2.testapp.databinding.FragmentAudiobookBinding
+import timber.log.Timber
+import kotlin.math.roundToLong
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 import kotlin.time.DurationUnit
 import kotlin.time.ExperimentalTime
 
-@OptIn(ExperimentalAudiobook::class, ExperimentalTime::class)
-class AudioReaderFragment : BaseReaderFragment() {
```
mickael-menu, line 27: I think we can keep it named `AudioReaderFragment` as we won't keep the media1 code in the test app.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        enum class State {
+            Playing,
+            Paused,
+            Finished,
+            Error
+        }
+    }
+
+    sealed class Exception
```
mickael-menu, line 97: What's the purpose of this empty `Exception` sealed class? It also inherits from `java.lang.Exception` instead of the Kotlin one.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Pauses the playback.
+     */
+    suspend fun pause(): Try<Unit, Exception>
+
+    /**
+     * Seeks to the given time at the given resource.
+     */
+
```
mickael-menu, line 37:
```suggestion
```
There are more newlines to remove in the same file.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
```diff
@@ -0,0 +1,99 @@
...
+     * Seeks to the beginning of the given link.
+     */
+
+    suspend fun go(link: Link): Try<Unit, Exception>
+
+    /**
+     * Seeks to the given locator.
+     */
+
+    suspend fun go(locator: Locator): Try<Unit, Exception>
```
mickael-menu, line 62: These are part of the `Navigator` interface already. I know they are less than ideal and not using coroutines, but this defeats the purpose of having a common interface (which is useful for shared screens, for example the table of contents or bookmarks). We can improve these APIs in Readium 3.x.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,274 @@
...
+    private suspend fun dummySeekBy(offset: Duration): Try<Unit, MediaNavigator.Exception> {
+        val newIndex = playerFacade.currentIndex!!
+        val newPosition = playerFacade.currentPosition!! + offset
+        return playerFacade.seekTo(newIndex, newPosition).toNavigatorResult()
+    }
+
+    override fun close() {
+        playerFacade.unregisterPlayerCallback(playerCallback)
+        playerCallback.close()
+        playerFacade.close()
```
mickael-menu, line 174: We probably need to cancel the coroutine scope there too?

```suggestion
        playerFacade.close()
        coroutineScope.cancel()
```

readium/navigator/src/main/java/org/readium/r2/navigator/media2/ExoPlayerFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+
+@ExperimentalAudiobook
+@OptIn(ExperimentalTime::class)
+class ExoPlayerFactory(private val cache: Cache? = null) {
+
+    fun createPlayer(context: Context, publication: Publication): SessionPlayer {
+        val dataSourceFactory = createDataSource(publication, cache)
+        val player: ExoPlayer = ExoPlayer.Builder(context)
+            .setSeekBackIncrementMs(30.seconds.inWholeMilliseconds)
+            .setSeekForwardIncrementMs(30.seconds.inWholeMilliseconds)
```
mickael-menu, line 27: Doesn't this conflict with the `MediaSessionNavigator.Configuration.skipForward/BackwardInterval` properties?

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.readium.r2.shared.util.Try
+import kotlin.time.Duration
+import kotlin.time.ExperimentalTime
+
+/**
+ * A navigator rendering an audio or video publication.
+ */
+@ExperimentalAudiobook
+@OptIn(ExperimentalTime::class)
+interface MediaNavigator : Navigator {
```
mickael-menu, line 17: I'm a bit on the fence for this interface. It doesn't seem to bring much compared to the existing `MediaNavigator` interface and the whole purpose was to have a single interface for various media-based navigators. Having one interface per implementation defeats the purpose.

I get that it adds coroutine support, but I'd rather update the existing interface and media1 implementation to use coroutines (even if it's only in the API signatures and not really in the implementation). It still has the `@ExperimentalAudiobook` annotation so it's fine to make breaking changes.

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
```diff
@@ -0,0 +1,136 @@
...
+            val activityIntent = createSessionActivityIntent(arguments)
+            mediaSession = navigator.session(applicationContext, arguments.bookId.toString(), activityIntent)
+                .also { addSession(it) }
+            mediaNavigator = navigator
+            saveLocationJob = navigator.currentLocator
+                .buffer(1, BufferOverflow.DROP_OLDEST)
+                .onEach {  locator ->
+                    delay(3.seconds)
+                    currentBookId?.let { id -> books.saveProgression(locator, id) }
+                }
```
mickael-menu, line 76: I think you can simplify this with a `debounce()`.

But I would try to keep the test app code as simple/dumb as possible and let apps decide if they need to do some optimisations. Also I guess the frequency can be customized with `Configuration.positionRefreshRate`?

```suggestion
                .debounce(3.seconds)
                .onEach {  locator ->
                    currentBookId?.let { id -> books.saveProgression(locator, id) }
                }
```

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
qnga: Mmm, yes. We often wait for garbage collection, but since there is a `close` method...

readium/navigator/src/main/java/org/readium/r2/navigator/media2/ExoPlayerFactory.kt
qnga: No, actually this is useless since the corresponding commands are not used. I'll remove it to prevent confusion.

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
qnga: No, `debounce` curiously didn't work here. My guess is that since all events are close enough each from other, they are all skipped.
`positionRefreshRate` should be kept high in order to refresh the time bar. Here we don't need such a high frequency. I think illustrating good practices is not bad.

test-app/src/main/java/org/readium/r2/testapp/reader/FullscreenReaderActivity.kt
```diff
@@ -16,10 +16,10 @@ import org.readium.r2.testapp.utils.showSystemUi
 /**
  * Adds fullscreen support to the ReaderActivity
  */
-open class VisualReaderActivity : ReaderActivity() {
+open class FullscreenReaderActivity : ReaderActivity() {
```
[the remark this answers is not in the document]
qnga: That doesn't look trivial to do and I'm not willing to tackle this right now.
I can merge again `FullscreenReaderFragment` and `VisualReaderFragment` if you prefer so.

test-app/src/main/java/org/readium/r2/testapp/utils/SystemUiManagement.kt
```diff
@@ -58,7 +59,7 @@ fun View.padSystemUi(insets: WindowInsets, activity: Activity) =
         .getInsets(WindowInsetsCompat.Type.systemBars()).apply {
             setPadding(
                 left,
-                top,
+                top + (activity as AppCompatActivity).supportActionBar!!.height,
```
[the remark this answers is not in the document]
qnga: Mmmm, the result would be wrong.  I'll complete the doc instead.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
qnga: NYPL doesn't use the common interface (so far at least), and I suppose we're not alone in doing so. It would be a pity to start a new navigator with technical debt, so I suggest we keep the matching methods public in  `MediaSessionNavigator` and remove them from the interface. Maybe that's what you were thinking about.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
qnga: I didn't think the two interfaces should coexist.  We are not in a case with two different media navigators, but with one supposed to replace the other one (still experimental). If breaking the interface is not a problem, let's just remove the previous navigator.

test-app/src/main/java/org/readium/r2/testapp/utils/SystemUiManagement.kt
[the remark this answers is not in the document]
mickael-menu: Sounds good

test-app/src/main/java/org/readium/r2/testapp/reader/FullscreenReaderActivity.kt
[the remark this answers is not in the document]
mickael-menu: Yeah in this case I'd prefer merging them and keeping `VisualReaderFragment` frontmost.

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
mickael-menu: I see, we would need `throttle` but it's not (yet) part of the coroutines library unfortunately. Let's keep your solution until it's available.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
mickael-menu: Are you not using the common interface because you have your own EPUB navigator? 

I'm really surprised you wouldn't use the share interface if you used more than one Readium navigator, for example it's used by the test app here: https://github.com/readium/kotlin-toolkit/blob/e2eeb4aea86d4d20726ebc71a33872a6c0bdb219/test-app/src/main/java/org/readium/r2/testapp/reader/BaseReaderFragment.kt#L115-L116

mickael-menu: But yeah, I'm fine if you remove the suspending versions from the interface but keep them in the implementation, for now.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
mickael-menu: Even though the media1 navigator is still marked as experimental, I still want to have a migration period before removing it entirely. That would also allow to merge this PR in even if it's not yet fully on par with the media1 navigator feature-wise and not yet battle-tested.

For the `MediaNavigator` interface, I'm considering that maybe we should just remove it. Contrarily to the other navigator interfaces, we will probably have only a single implementation in Readium, so it's not very useful. I suggest that we wait until we have another implementation (e.g. for video publications) in Readium or an open third-party audio navigator before extracting a shared interface.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,272 @@
...
+    private val configuration: Configuration
+) : MediaNavigator {
+
+    private val coroutineScope: CoroutineScope = MainScope()
+
+    private fun <T> Flow<T>.stateInFirst(coroutineScope: CoroutineScope): StateFlow<T> =
+        stateIn(coroutineScope, SharingStarted.Lazily, runBlocking { first() })
+
+    private val totalDuration: Duration? =
+        this.playerFacade.playlist!!.durations?.sum()
```
stevenzeck, line 48: Are we sure `playlist` and others here in this file coming from `playerFacade` won't ever be null?

readium/navigator/src/main/java/org/readium/r2/navigator/media2/SessionPlayerHelpers.kt
```diff
@@ -0,0 +1,105 @@
...
+        ms.milliseconds
+
+/**
+ * Metadata
+ */
+
+@ExperimentalAudiobook
+internal val MediaMetadata.href: String
+    get() = getString(MediaMetadata.METADATA_KEY_MEDIA_URI)
+        .let { checkNotNull(it) { "Missing href in item metadata."} }
```
stevenzeck, line 42: Shouldn't need `let` here

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
stevenzeck: `sample` should work though.  Replace `debounce` with `sample` in @mickael-menu's suggested change.

test-app/src/main/java/org/readium/r2/testapp/utils/LifecycleMediaSessionService.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    @CallSuper
+    override fun onStart(intent: Intent?, startId: Int) {
+        lifecycleDispatcher.onServicePreSuperOnStart()
+        super.onStart(intent, startId)
+    }
+
+    // this method is added only to annotate it with @CallSuper.
+    // In usual service super.onStartCommand is no-op, but in LifecycleService
+    // it results in mDispatcher.onServicePreSuperOnStart() call, because
```
stevenzeck, line 40: But I don't think `onStart` is ever called in this case, thus `lifecycleDispatcher.onServicePreSuperOnStart()` is never called above.

qnga: Why do you think `onStart` is never called?
I mostly copy-pasted this code from `LifeycleService` (and should have written so):
https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-main/lifecycle/lifecycle-service/src/main/java/androidx/lifecycle/LifecycleService.java

readium/navigator/src/main/java/org/readium/r2/navigator/media2/SessionPlayerHelpers.kt
qnga: That looks clearer to me so.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
qnga: Yes, I am.  `MediaSessionNavigator` is created only once the `SessionPlayer` behind is ready.

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
qnga: Will try, thanks!

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
qnga: We have one custom `Viewer` interface which is used by the `ServiceLoader` to dynamically load all available `Viewer` on the class path. This enables us to have a different set of viewers in different apps. Thus, we used to have each viewer in a separate module.
I think the benefits of the use of the common interface with both visual and audio navigators is quite tight. It's likely to save a couple of lines of code.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
qnga: Mmm, that's a legacy. It shouldn't be sealed any more, and it looks like the compiler treats it as not sealed.
Will switch to the Kotlin class.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
qnga: Agreed.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
qnga: Oh no! sealed class children can be declared anywhere in the package now (I think they used to have to be in the same file). So that worked well.

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
mickael-menu: Nice, I didn't know about `sample`. Looks like a `throttleLast`.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/SessionPlayerFacade.kt
```diff
@@ -0,0 +1,200 @@
...
+        sessionPlayer.unregisterPlayerCallback(callback)
+    }
+
+    fun close() {
+        sessionPlayer.close()
+    }
+
+    fun session(context: Context, id: String, activityIntent: PendingIntent): MediaSession {
+        return MediaSession.Builder(context, sessionPlayer)
+            .setId(id)
```
mickael-menu, line 67: According to the doc:

> Use this **if and only if** your app supports multiple playback at the same time and also wants to provide external apps to have finer controls of them.
> https://developer.android.com/reference/androidx/media2/session/MediaSession.Builder#setId(java.lang.String)

So maybe we should use `id: String? = null` and set it only if non-null.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/SessionPlayerFacade.kt
```diff
@@ -0,0 +1,200 @@
...
+                    seekToSync(position)
+                }
+                position == Duration.ZERO -> {
+                    skipToPlaylistItemSync(index)
+                }
+                else -> {
+                    skipToPlaylistItemSync(index)
+                        // WORKAROUND: skipToPlayListItem terminates before Exoplayer is aware that
+                        // the new item is seekable. Not to wait would result in the command being skipped.
+                        .also { Thread.sleep(500) }
```
mickael-menu, line 108: Isn't there a way to make this a bit less fragile? There's no callback event when the current item changes that could be used?

readium/navigator/src/main/java/org/readium/r2/navigator/media2/SessionPlayerResult.kt
```diff
@@ -0,0 +1,49 @@
...
+    SESSION_PARENTAL_CONTROL_RESTRICTED,
+    SESSION_PREMIUM_ACCOUNT_REQUIRED,
+    ERROR_SESSION_SETUP_REQUIRED,
+    SESSION_SKIP_LIMIT_REACHED,
+    UNKNOWN,
+    INFO_SKIPPED;
+
+    companion object {
+
+        fun fromSessionResultCode(resultCode: Int): SessionPlayerError {
```
mickael-menu, line 28: Meh, I hope they fixed these error codes in `media3`.

test-app/build.gradle.kts
```diff
@@ -100,6 +100,9 @@ dependencies {
     // AM NOTE: needs to stay this version for now (June 24,2020)
     implementation("org.jsoup:jsoup:1.14.3")
 
+    implementation("androidx.media2:media2-session:1.2.0")
+    implementation("androidx.media2:media2-player:1.2.0")
```
mickael-menu, line 104: You can use `api` instead of `implementation` in `navigator-media/build.gradle` to make the `media2` types automatically visible in the Test App when importing `readium-navigator-media`, without requiring to import them explicitly.

test-app/src/main/java/org/readium/r2/testapp/utils/ViewLifecycleDelegates.kt
```diff
@@ -0,0 +1,75 @@
...
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
+
+/**
+ * Repository for values automatically set to null every time the [Fragment]'s view is destroyed.
+ *
+ * This is especially useful for binding properties to avoid memory leaks when Fragments are on
+ * the back stack without any view attached.
+ */
+class LifecycleDelegates(private val fragment: Fragment):  DefaultLifecycleObserver {
```
mickael-menu, line 23: Nice!

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
```diff
@@ -0,0 +1,123 @@
...
+import org.readium.r2.navigator.media2.MediaSessionNavigator
+import org.readium.r2.testapp.bookshelf.BookRepository
+import org.readium.r2.testapp.db.BookDatabase
+import org.readium.r2.testapp.reader.ReaderActivityContract
+import org.readium.r2.testapp.utils.LifecycleMediaSessionService
+import timber.log.Timber
+import kotlin.time.ExperimentalTime
+
+@OptIn(ExperimentalTime::class, ExperimentalAudiobook::class, ExperimentalCoroutinesApi::class)
+class MediaService : LifecycleMediaSessionService() {
```
mickael-menu, line 29: This code might be a bit dense for newcomers, could you add some comments to describe it? That's important if they copy-paste it in their apps.

test-app/src/main/java/org/readium/r2/testapp/reader/NavigatorType.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2022 Readium Foundation. All rights reserved.
+ * Use of this source code is governed by the BSD-style license
+ * available in the top-level LICENSE file of the project.
+ */
+
+package org.readium.r2.testapp.reader
+
+enum class NavigatorType {
```
mickael-menu, line 9: This doesn't seem to be used anymore.

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt
```diff
@@ -34,50 +36,46 @@ import org.readium.r2.testapp.outline.OutlineFragment
...
+        modelFactory = ReaderViewModel.Factory(app, arguments)
+        model = ViewModelProvider(this)[ReaderViewModel::class.java]
+
+        /*
+         * [ReaderViewModel.Factory] provides dummy publications if the [ReaderActivity] is restored
+         * after the app process was killed because the [ReaderRepository] is empty.
+         * In that case, finish the activity as soon as possible and go back to the previous one.
+         */
+        if (model.publication.readingOrder.isEmpty()) {
+            finish()
```
mickael-menu, line 56: Some navigator fragments will crash there if we continue with an empty `readingOrder`.
```suggestion
            finish()
            super.onCreate(savedInstanceState)
            setContentView(View())
            return
```

mickael-menu (changes requested):

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+    override val publication: Publication,
+    private val playerFacade: SessionPlayerFacade,
+    private val playerCallback: SessionPlayerCallback,
+    private val configuration: Configuration
+) : Navigator {
+
+    private val coroutineScope: CoroutineScope = MainScope()
+
+    private fun <T> Flow<T>.stateInFirst(coroutineScope: CoroutineScope): StateFlow<T> =
+        stateIn(coroutineScope, SharingStarted.Lazily, runBlocking { first() })
```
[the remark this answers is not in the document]
qnga: Neither of them. `runBlocking` is here only in order to get rid of `suspend`. By design, the first element is necessarily already available when the constructor is called.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+
+    private val totalDuration: Duration? =
```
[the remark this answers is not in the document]
qnga: Those durations are got by summing the `Link` durations. Is that clear to you?
I'm not sure about exposing something not especially reliable and may be missing.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+            playerCallback.playbackSpeed,
+            playerCallback.currentItem,
+            playerCallback.currentPosition,
+            playerCallback.bufferedPosition
+        ) { currentState, playbackSpeed, currentItem, currentPosition, bufferedPosition ->
+            val state = when (currentState) {
+                SessionPlayerState.Playing ->
+                    Playback.State.Playing
+                SessionPlayerState.Idle, SessionPlayerState.Error ->
+                    Playback.State.Error
```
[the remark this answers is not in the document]
qnga: The `MediaSessionNaviator` is created only after the playlist has been set up and the player state is supposed to have reached `Paused`. That's the design's key.

test-app/build.gradle.kts
qnga: Yes, I can. Should I? It might be a better practice to let modules be honest about which modules they depend on.

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
qnga: Which code?

test-app/src/main/java/org/readium/r2/testapp/reader/NavigatorType.kt
qnga: Oh year, thanks!

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt
qnga: Isn't it exactly the way it used to work with intent helpers?

readium/navigator/src/main/java/org/readium/r2/navigator/media2/SessionPlayerFacade.kt
qnga: I can't see the documentation's point against IDs. I used the ID before, now not any more so I can remove the ID session.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/SessionPlayerFacade.kt
qnga: No, I think that's a bug. Looking at the code, if I remember correctly, I supposed the callback would be called too early as well.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+                // Player must be ready to play when MediaNavigator's constructor is called.
+                .map { MediaSessionNavigator(publication, facade, callback, configuration) }
+        }
+
+        private suspend fun preparePlayer(
+            publication: Publication, player:
+            SessionPlayerFacade
+        ): Try<Unit, Exception> {
+            val playlist = publication.readingOrder.toPlayList()
+            val metadata = publicationMetadata(publication)
```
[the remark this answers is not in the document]
qnga: I don't know. Are media style notifications limited to player's metadata? If not, apps are free to use publication's metadata when they build up their notifications.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+
+        enum class State {
+            Playing,
+            Paused,
+            Finished,
+            Error
+        }
+    }
+
+    sealed class Exception : kotlin.Exception()
```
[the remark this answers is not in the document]
qnga: The player doesn't provide exceptions, I can't make them up....

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+        val currentPosition: Duration,
+        val bufferedPosition: Duration,
+    ) {
+
+        enum class State {
+            Playing,
+            Paused,
+            Finished,
+            Error
+        }
```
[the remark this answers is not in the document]
qnga: `Idle` -> `Paused` is used.
`Loading` -> Presumably, the global state would be `Paused` or `Playing` which makes sense, I think.  Media2 has a special buffer state which I didn't explore. I'll have a look.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+                durations
+            )
+        Timber.d("Smart seeking by $offset resolved to item $newIndex position $newPosition")
+        return playerFacade.seekTo(newIndex, newPosition).toNavigatorResult()
+    }
+
+    private suspend fun dummySeekBy(offset: Duration): Try<Unit, Exception> {
+        val newIndex = playerFacade.currentIndex!!
+        val newPosition = playerFacade.currentPosition!! + offset
+        return playerFacade.seekTo(newIndex, newPosition).toNavigatorResult()
```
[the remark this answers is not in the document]
qnga: Good question.
See SessionPlayerFacade line 179.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
stevenzeck: This looks to be the default one, as mentioned [here](https://developer.android.com/guide/topics/media/media-controls#mediasession_implementation).  I'm wondering whether it has a default at all for Android 10 and below though.

But otherwise it would be the developer's responsibility to override `onUpdateNotification` when extending `MediaSessionService` in their app to display their own notification control.  However, the metadata in that notification needs to keep in sync with what is in the MediaSession in the Navigator, and we should ensure that is possible.  We should also ensure developer's can control aspects of playback so they can add their own actions in the notification controls.

Since the metadata we inject in the Navigator is also accessible from `MediaSession` being passed into `onUpdateNotification`, I would suggest we at least add the following:

- METADATA_KEY_ARTIST (a default):  Author(s) of the publication
- METADATA_KEY_TITLE (a default):  Title of the publication
- METADATA_KEY_ALBUM_ART:  Bitmap of the publication cover
- METADATA_KEY_DISPLAY_SUBTITLE:  The title of the current chapter

And consider adding:

- METADATA_KEY_NUM_TRACKS:  Number of chapters of the publication
- METADATA_KEY_AUTHOR:  Author(s) of the publication

readium/navigator/src/main/java/org/readium/r2/navigator/media2/ExoPlayerFactory.kt
```diff
@@ -0,0 +1,51 @@
...
+    private fun createDataSource(publication: Publication, cache: Cache?): DataSource.Factory {
+        var factory: DataSource.Factory = PublicationDataSource.Factory(publication)
+
+        if (cache != null) {
+            factory = CacheDataSource.Factory()
+                .setCache(cache)
+                .setUpstreamDataSourceFactory(factory)
+                // Disable writing to the cache by the player. We'll handle downloads through the
+                // service.
+                .setCacheWriteDataSinkFactory(null)
```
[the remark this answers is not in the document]
qnga: What about
```
class ExoPlayerFactory(
    private val onCreateExoPlayer: (ExoPlayer.Builder).() -> Unit,
    private val onCreateDataSource: (DataSource.Factory).() -> Unit 
)

```
?

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+    val playback: StateFlow<Playback> =
+        playbackStateFlow.stateInFirst(coroutineScope)
+
+    /**
+     * Sets the speed of the media playback.
+     *
+     * Normal speed is 1.0 and 0.0 is incorrect.
+     */
+    suspend fun setPlaybackRate(rate: Double): Try<Unit, Exception> =
+        playerFacade.setPlaybackSpeed(rate).toNavigatorResult()
```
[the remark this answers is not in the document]
qnga: I'm not sure. My purpose was to transparently pass commands to the player, which would accept or reject them. For instance, `SessionPlayer` supports reverse playback with negative rates. I don't know if that may be useful to anyone.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+     */
+    fun close() {
+        playerFacade.unregisterPlayerCallback(playerCallback)
+        playerCallback.close()
+        playerFacade.close()
+        coroutineScope.cancel()
+    }
+
+    fun session(context: Context, id: String, activityIntent: PendingIntent): MediaSession =
+        playerFacade.session(context, id, activityIntent)
```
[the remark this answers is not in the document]
qnga: I think so.
This method is explained in the class header.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/SessionPlayerFacade.kt
qnga: Done.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+     */
+    suspend fun seek(index: Int, position: Duration): Try<Unit, Exception> =
+        playerFacade.seekTo(index, position).toNavigatorResult()
+
+    /**
+     * Seeks to the given locator.
+     */
+    suspend fun go(locator: Locator): Try<Unit, Exception> {
+        Timber.d("Go to locator $locator")
+        val itemIndex = checkNotNull(publication.readingOrder.indexOfFirstWithHref(locator.href))
```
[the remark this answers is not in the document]
qnga: Fixed.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
```diff
@@ -0,0 +1,327 @@
...
+        val skipBackwardInterval: Duration = 30.seconds,
+    )
+
+    @ExperimentalTime
+    data class Playback(
+        val state: State,
+        val rate: Double,
+        val currentIndex: Int,
+        val currentLink: Link,
+        val currentPosition: Duration,
```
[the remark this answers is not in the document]
qnga: Done.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
qnga: Done.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaNavigator.kt
mickael-menu: Indeed, I didn't know that.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: Okay

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: It depends if using a negative rate crashes the navigator or not, in which case it becomes a precondition of the navigator IMO. Allowing this also implies that this is a feature of the navigator and it becomes a breaking change to remove it at some point if we see that it doesn't work well with the navigator.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: Okay

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: No, I agree it's not very useful in this case.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/ExoPlayerFactory.kt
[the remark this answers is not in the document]
mickael-menu: Yes, okay. Especially since using `setWakeMode()` also requires to add a permission in the Android manifest.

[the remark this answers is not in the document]
mickael-menu: I don't think this was used in media1 so maybe it's not really useful. Let's see if we really need it, in which case this should be documented in the test app's background service.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: Yeah for `Loading` in case of buffering that would be important to display a spinner in the user interface. Otherwise you get this feeling of something broken.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: Yes we don't need one exception per media3 error code, but we could use a `PlayerException` that wraps the raw code, or another one when the user provides an invalid locator to `go()`, or another wrapping `PublicationDataSource.Exception` if we get it at some point.

Regarding the usefulness, as a toolkit we need to provide as much details as possible for the (nested) error causes to help implementers find out the root cause of an issue and choose whether (or not) and how to recover from it, depending on their strategy.

I recommend this great article about the difference between app-level and library-level error handling: https://nick.groenen.me/posts/rust-error-handling/#libraries-versus-applications

> It took me a moment to appreciate this distinction, but there’s value in approaching error handling differently between libraries and applications as they tend to have different concerns:
> * Libraries should focus on producing meaningful, structured error types/variants. This allows applications to easily differentiate various error cases.
> * Applications mainly consume errors.
> * Libraries may want to cast errors from one type to another. An IO error should likely be wrapped by a high-level error type provided by the library.
>      * Otherwise an IO error in library `foo` cannot be distinguished from a similar IO error in library `bar`.
> * Where libraries return errors, applications decide if and how those errors are formatted and displayed to users.
> * Applications may also want to parse and inspect errors, for example to forward them to exception tracking services or to retry operations when doing so is deemed to be safe.

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
mickael-menu: * The purpose of `createSessionActivityIntent()`.
* Why we save the `currentLocator` in the DB in the service instead of the `Activity`.
* Why we close the navigator in `onTaskRemoved()`.

I know implementers can RTFM but since this service is likely to be copy-pasted, I think it could help to have a bit more explanations for people to understand what they have in their apps.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: I'll try it out when I'll test the PR again

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt
mickael-menu: The intent helpers couldn't act at this level, but this was supposed to be the "correct" way to use it in the test app. It's possible that it wasn't implemented properly before.

For example in `PdfNavigatorFragment`, you have this in the `onCreateView`:

```
go(publication.readingOrder.first())
```

Also even with this suggestion I'm not sure this is 100% crash-free with the fragments recovery. This part is really shaky like we discussed, hopefully we can implement a better solution by using maybe Navigation component at some point.

test-app/build.gradle.kts
mickael-menu: IMO, a good rule of thumb is that if you expose a type from a nested dependency in your public surface API, then effectively the nested dependency is now part of your own public API and you _should_ use `api` (might be why they chose this keyword).

However, I don't have a strong opinion for this case as the type of `media2` that are exposed are not required to be used in the test app for the navigator to work, so an explicit opt-in seems fine.

Anyways, I tried commenting these two `implementation` statements and even without the use of `api` the test app still builds and runs fine (I cleared all the cache). This is weird no?

mickael-menu: Ha, I think it's because we use `api` for ExoPlayer in the `navigator` module: 
https://github.com/readium/kotlin-toolkit/blob/72fa156d85173c186bb0f2aae020be2122c6abc2/readium/navigator/build.gradle.kts#L90-L94

And ExoPlayer itself is using `api` for the `media2` dependency (per the rule of thumb above), so transitively this is accessible in the Test App.

https://github.com/google/ExoPlayer/blob/029a2b27cbdc27cf9d51d4a73ebeb503968849f6/extensions/media2/build.gradle#L20

For now you can leave it like this and we'll adjust when removing the older media1 navigator.

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt
qnga: I think the intent helper acts here because of this line https://github.com/readium/kotlin-toolkit/blob/adeb8f452fbd48b6ab7e1132264bcce9ab64ee95/test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt#L47
This is mostly like the new code. The Activity is told to finish, but it doesn't do it immediately. The rest of the code is executed with a publication with an empty reading order, and an exception might be thrown https://github.com/readium/kotlin-toolkit/blob/adeb8f452fbd48b6ab7e1132264bcce9ab64ee95/test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt#L74
Anyway, I (hopefully) made the `ReaderActivity` crash-free yesterday.

test-app/src/main/java/org/readium/r2/testapp/MediaService.kt
qnga: I had to read the fucking manual several times and do many experiments to get this stuff working.
Done.

test-app/build.gradle.kts
qnga: Ok

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
qnga: I can't see any reason for a negative rate to crash the navigator or cause any issue. I think the playback speed is a very specific setting which doesn't have any consequences anywhere.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
qnga: As a fan of Rust, I knew this article! It reminds me that I'm not sure about  the attempts to provide error messages in the toolkit (`UserException`).

About `PublicationDataSource.Exception`, I feel like the `SessionPlayer` would go into the error state if any occur, and that's it. Let me know if you have more insights.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: I'll try it out quickly on Friday

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: > It reminds me that I'm not sure about the attempts to provide error messages in the toolkit (`UserException`).

That's fair, I started this for the LCP errors which can be quite complicated and require some specific mentions in the user messages to pass the certification.

But I agree that returning `UserException` should not be required (or maybe even recommended) for other APIs. As long as we have properly typed and detailed ones and don't return `kotlin.Exception`.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
qnga: I've added something for this.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: It works fine, the offset is truncated to the available duration.

readium/navigator/src/main/java/org/readium/r2/navigator/media2/MediaSessionNavigator.kt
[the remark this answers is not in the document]
mickael-menu: I can't test this out, ExoPlayer explicitly prevents negative playback speed:

https://github.com/google/ExoPlayer/blob/029a2b27cbdc27cf9d51d4a73ebeb503968849f6/extensions/media2/src/main/java/com/google/android/exoplayer2/ext/media2/SessionPlayerConnector.java#L143

It doesn't return any error as ExoPlayer assertions are not enabled, the command is just ignored.

Let's leave it at that for now.

mickael-menu (approved):

PR #81, opened 2022-01-13
Fix invisible `<audio>` elements

mickael-menu:
#### Navigator

* [#193](https://github.com/readium/r2-navigator-kotlin/issues/193) Fixed invisible `<audio>` elements.


qnga (approved):
Looks good, as far as I can tell.

PR #83, opened 2022-01-21
Deprecate `Link.toLocator()` in favor of `Publication.locatorFromLink()`

mickael-menu:
#### Shared

* `Link.toLocator()` is deprecated as it may create an incorrect `Locator` if the link `type` is missing.
    * Use `publication.locatorFromLink()` instead.

---

See Slack discussion for context: https://readium.slack.com/archives/C703MSTQU/p1642520174001200

### DIFF
--- a/readium/navigator-media2/src/main/java/org/readium/navigator/media2/MediaNavigator.kt
+++ b/readium/navigator-media2/src/main/java/org/readium/navigator/media2/MediaNavigator.kt
@@ -82,7 +82,8 @@ class MediaNavigator private constructor(
             if (itemStartPosition == null) null
             else totalDuration?.let { (itemStartPosition + position) / it }
 
-        return link.toLocator().copyWithLocations(
+        val locator = requireNotNull(publication.locatorFromLink(link))
+        return locator.copyWithLocations(
             fragments = listOf("t=${position.inWholeSeconds}"),
             progression = item.duration?.let { position / it },
             totalProgression = totalProgression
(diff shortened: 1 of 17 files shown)


readium/shared/src/main/java/org/readium/r2/shared/publication/Manifest.kt
```diff
@@ -64,6 +64,38 @@ data class Manifest(
...
+    /**
+     * Finds the first [Link] with the given HREF in the manifest's links.
+     *
+     * Searches through (in order) [readingOrder], [resources] and [links] recursively following
+     * alternate and children links.
+     *
+     * If there's no match, try again after removing any query parameter and anchor from the
+     * given [href].
+     */
+    fun linkWithHref(href: String): Link? {
```
mickael-menu, line 76: I moved this from `Publication`, as it makes sense to have it in `Manifest` as well.

test-app/src/main/java/org/readium/r2/testapp/outline/NavigationFragment.kt
```diff
@@ -83,13 +83,7 @@ class NavigationFragment : Fragment() {
     }
 
     private fun onLinkSelected(link: Link) {
-        val locator = link.toLocator().let {
-            // progression is mandatory in some contexts
-            if (it.locations.fragments.isEmpty())
-                it.copyWithLocations(progression = 0.0)
-            else
-                it
```
mickael-menu, line 91: I made sure that `locatorFromLink()` returns a `Locator` with either `progression` or `fragments` set.

readium/navigator-media2/src/main/java/org/readium/navigator/media2/MediaNavigator.kt
```diff
@@ -274,16 +278,15 @@ class MediaNavigator private constructor(
...
+            internal val error: SessionPlayerError
+        ) : Exception("${error.name} error occurred in SessionPlayer.")
+
+        class IllegalArgument(message: String): Exception(message)
 
-        class IllegalArgument internal constructor(
-            override val message: String
-        ): Exception()
+        class ResourceNotFound(val href: String) : Exception("Resource not found at $href")
     }
```
qnga, line 290: `IllegalArgument` is used for the exact same purpose (stale link or locator). I think it's good enough, though it might be better named `InvalidArgument`.

qnga (changes requested):

qnga (approved):
Looks good.

PR #84, opened 2022-01-27
Update Android Gradle Plugin to 7.1.0

stevenzeck:
This PR does a few minor things.

1. Update the Android Gradle Plugin to 7.1.0
2. Following the release of Android Studio 2021.1.1, when creating a new project the gradle file contents have changed (see https://developer.android.com/studio/releases/gradle-plugin#settings-gradle). This PR makes those same changes.
    1. Repository settings moved from top level gradle to settings.gradle.kts
    2. The top level gradle file only contains the plugins block and gradle tasks
    3. The dependencies that were in that top level gradle seem to be replaced by plugins. `com.android.tools.build:gradle` replaced by `com.android.application` and `com.android.library`, `org.jetbrains.kotlin:kotlin-gradle-plugin` replaced by `org.jetbrains.kotlin.android`, `org.jetbrains.dokka:dokka-gradle-plugin` replaced by `org.jetbrains.dokka`.  Sadly the plugins block doesn't seem to let you use variables for the version numbers
3. Update other dependencies throughout the modules, one of which required changing `runBlockingTest` to `runTest`
4. Fixed warnings, including removing parameters not being used as well as name shadowing


mickael-menu:
There seems to be an issue with the JitPack generation. It builds successfully but the artifacts are not created. You can find the logs here: https://jitpack.io/com/github/stevenzeck/kotlin-toolkit/agp-upgrade-8f47ee0e18-1/build.log

mickael-menu (approved):
The issue seems to have been from JitPack's side, I was able to run the built artifacts from `r2-testapp-kotlin`, `gradle6.9` branch. Looking good to me, thanks!

PR #85, opened 2022-02-15
Fix selection of Epub unique identifier

### LINKED ISSUE #70: Incorrect parsing of the EPUB `dc:identifier`
According to the [parser specification](https://github.com/readium/architecture/blob/master/streamer/parser/metadata.md#epub-3x-1):

> The primary `identifier` is marked as the unique identifier via the `package` element `unique-identifier` attribute. It’s `id` must therefore match the value of this `unique-identifier` attribute.

But it looks like our parser is taking the last `dc:identifier` element found, which in this sample is `2016-11-05`. It should be `urn:uuid:65784686-a461-41a7-8c5c-5cd26d7b7b8d`:

```xml
<?xml version='1.0' encoding='utf-8'?>
<package xmlns="http://www.idpf.org/2007/opf" version="2.0" unique-identifier="uuid_id">
  <metadata xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:opf="http://www.idpf.org/2007/opf" xmlns:dcterms="http://purl.org/dc/terms/"  xmlns:dc="http://purl.org/dc/elements/1.1/">
    <dc:title>Germaine</dc:title>
    <dc:identifier opf:scheme="filename">about_edmond_-_germaine</dc:identifier>
    <dc:identifier opf:scheme="ISBN">978-2-8247-1198-0</dc:identifier>
    <dc:identifier id="BookId" opf:scheme="URI">http://www.bibebook.com/search/978-2-8247-1198-0</dc:identifier>
    <dc:identifier opf:scheme="SOURCE">BEQ</dc:identifier>
    <dc:identifier opf:scheme="NEXT"></dc:identifier>
    <dc:identifier id="uuid_id" opf:scheme="UUID">urn:uuid:65784686-a461-41a7-8c5c-5cd26d7b7b8d</dc:identifier>
    <dc:identifier opf:scheme="version">2016-11-05</dc:identifier>
  </metadata>
</package>
```

qnga:
Fix #70. Sorry about that, silly mistake and too weak tests.


mickael-menu (approved):
Looking good, thank you for addressing it.

PR #95, opened 2022-03-23
Use `startForegroundService()` for the `MediaService`

### LINKED ISSUE #92: Caused by java.lang.IllegalStateException: Not allowed to start service Intent
**Device**
Brand:samsung
Model:Galaxy S9
Orientation: Portrait
RAM free: 1.2 GB
Disk free: 14.3 GB

**Operating System**
Version:Android 10
Orientation: Portrait
Rooted:No

```
Caused by java.lang.IllegalStateException: Not allowed to start service Intent { act=com.readerappnext.MediaService cmp=com.readerappnext/.epub_plugin.MediaService }: app is in background uid UidRecord{3c87fd2 u0a230 TRNB idle change:uncached procs:1 proclist:26821, seq(0,0,0)}
       at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1688)
       at android.app.ContextImpl.startService(ContextImpl.java:1633)
       at android.content.ContextWrapper.startService(ContextWrapper.java:683)
       at com.theonelibrary.Application.onCreate(Application.java:74)
       at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1190)
       at android.app.ActivityThread.handleMakeApplication(ActivityThread.java:7184)
       at android.app.ActivityThread.handleBindApplication(ActivityThread.java:7134)
       at android.app.ActivityThread.access$1600(ActivityThread.java:274)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2102)
       at android.os.Handler.dispatchMessage(Handler.java:107)
       at android.os.Looper.loop(Looper.java:237)
       at android.app.ActivityThread.main(ActivityThread.java:8167)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:496)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1100)
```

running latest develop code. Error comes occasionally not every time.

mickael-menu:
Fix #92

Solution found in https://stackoverflow.com/a/47654126

### DIFF
--- a/readium/shared/src/main/java/org/readium/r2/shared/FuelPromiseExtension.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/FuelPromiseExtension.kt
@@ -16,7 +16,7 @@ import nl.komponents.kovenant.Promise
 import nl.komponents.kovenant.deferred
 import nl.komponents.kovenant.task
 
-@Deprecated("Dependency to Fuel will eventually be removed from the Readium Toolkit")
+@Deprecated("Dependency to Fuel and kovenant will eventually be removed from the Readium Toolkit")
 fun Request.promise(): Promise<Triple<Request, Response, ByteArray>, Exception> {
     val deferred = deferred<Triple<Request, Response, ByteArray>, Exception>()
     task { response() } success {
--- a/readium/shared/src/main/java/org/readium/r2/shared/FuelPromiseExtension.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/FuelPromiseExtension.kt
@@ -16,6 +16,7 @@ import nl.komponents.kovenant.Promise
 import nl.komponents.kovenant.deferred
 import nl.komponents.kovenant.task
 
+@Deprecated("Dependency to Fuel will eventually be removed from the Readium Toolkit")
 fun Request.promise(): Promise<Triple<Request, Response, ByteArray>, Exception> {
     val deferred = deferred<Triple<Request, Response, ByteArray>, Exception>()
     task { response() } success {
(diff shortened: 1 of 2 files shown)


qnga (approved):
I believe you! Any chance there is also something to do in https://github.com/readium/kotlin-toolkit/blob/develop/test-app/src/main/java/org/readium/r2/testapp/MediaService.kt?

mickael-menu:
@qnga I guess it would be here: https://github.com/readium/kotlin-toolkit/blob/5234c25e2a5e922e0373f743034256f7d18597c7/test-app/src/main/java/org/readium/r2/testapp/Application.kt#L85

But I don't think using `startForegroundService()` here would be helpful, because I guess it will prevent the app from being destroyed while in background as it is started on app launch.

Like with the previous audio navigator, the media service should be started only when actually playing an audiobook, and stopped when we're done. It needs to be started as "foreground" for the issue raised in #92  but also this:

> You use [startForeground(int, Notification)](https://developer.android.com/reference/android/app/Service#startForeground(int,%20android.app.Notification)) if killing your service would be disruptive to the user, such as if your service is performing background music playback, so the user would notice if their music stopped playing.
>
> https://developer.android.com/reference/android/app/Service#startForeground(int,%20android.app.Notification)

qnga:
Correct, the service is made foreground only when some audiobook is opened. Let's assume that `media2` does all the work correctly until proven otherwise.

mickael-menu:
Fine by me.

@mrifni Did you have time to test this PR?

readium/shared/src/main/java/org/readium/r2/shared/FuelPromiseExtension.kt
```diff
@@ -16,6 +16,7 @@ import nl.komponents.kovenant.Promise
 import nl.komponents.kovenant.deferred
 import nl.komponents.kovenant.task
 
+@Deprecated("Dependency to Fuel will eventually be removed from the Readium Toolkit")
```
stevenzeck, line 19: @mickael-menu dependency to both Fuel and kovenant promise/deferred/task

mrifni:
@mickael-menu this seems to work, can we merge it ?

PR #101, opened 2022-04-06
Allow to disable display cutouts padding

szymn:
We have a situation where:
1. we embed the navigator (not full screen)
2. our host activity already handles insets and we want it to keep using LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES

In this scenario we want to avoid additional padding applied for insets within the reader.


404-html:
@mickael-menu @qnga @stevenzeck is there any chance for this being merged anytime soon 🙏

mickael-menu (approved):
@szymn Thank you! I took this opportunity for a small refactoring, the `R2BasicWebView` should not have a direct reference to the navigator.

PR #112, opened 2022-04-27
Dependency updates

stevenzeck:
1. Targets Android SDK 32, aka Android 12L
2. `-Xopt-in` is deprecated and replaced with `-opt-in`
3. Android Gradle plugin updated to 7.1.3
4. Kotlin updated to 1.6.21
5. All library dependencies updated in both `.gradle.kts` and the `.toml` files
6. The one code change is related to https://youtrack.jetbrains.com/issue/KT-47902.  When overriding deprecated methods, those overrides themselves have to be marked as deprecated.  Since `PublicationContainer` extends `Container` and overrides `data`, `dataLength`, and `dataInputStream` (all of which are deprecated), those have to marked as deprecated in PublicationContainer as well.

Note:  Gradle itself is still on 7.2, while can go to 7.4.2.


mickael-menu (approved):
Thank you!

PR #118, opened 2022-05-02
Text-to-speech and content extraction

### LINKED ISSUE #68: TTS support
Although the testapp currently has TTS, it is all bundled in the testapp module itself.  This means that developers are required to write their own TTS implementation with little to no support from the Readium libraries themselves.

mickael-menu:
#### Shared

* Extract the raw content (text, images, etc.) of a publication. [Take a look at the user guide](https://github.com/readium/kotlin-toolkit/blob/feature/tts/docs/guides/content.md).

#### Navigator

* A brand new text-to-speech implementation. [Take a look at the user guide](https://github.com/readium/kotlin-toolkit/blob/feature/tts/docs/guides/tts.md).

## Review notes

A good first step is to read the two user guides linked in the changelog above.

Then I suggest going in this order:

* Implementation of the [Content Iterator proposal](https://github.com/readium/architecture/blob/proposal/content-iterator/proposals/000-content-iterator.md) (with some modifications).
    * `Content.kt` contains the `Content` models as well as the `Content.Iterator` interface.
    * `PublicationContentIterator.kt` is a general implementation to iterate over a full publication, it delegates the actual extraction to `ResourceContentIterator` instances.
    * `HtmlResourceContentIterator` is a specialized `Content.Iterator` which can extract the content from an HTML `Resource`.
* `shared/util/tokenizer` is a new package providing utilities to tokenize pieces of data.
    * `TextTokenizer.kt` is used to split a `String` into multiple `IntRange` components. You can split a paragraph into sentences with it.
    * `ContentTokenizer.kt` splits a single `Content.Element` into a list of `Content.Element`.
        * The default `TextContentTokenizer` implementation delegates the actual text tokenizing to a provided `TextTokenizer` to split `Content.TextElement` into multiple `Content.TextElement`. Then, it will split the `Locator.Text` model to match the generated tokens.
* `Language.kt` is basically `Locale` but (ultimately) independent from the JVM.
* `EpubNavigator` was updated to not refresh the current resource when jumping to a locator in it. The TTS jumps to locator every seconds so it was causing a lot of visual refreshes. 
* `navigator/tts` package.
    * Note that the TTS is completely independent from the Navigator, we could put it in `shared` or even a new `readium-tts` module. However I put it in `navigator` as it's usually used in combination with a Navigator. We could raise this discussion in a call.
    * `TtsEngine` is an interface to implement to support a third-party TTS engine. It is pretty simple with a `speak()` API to speak a single utterance. It's suspending so cancelling the job cancels the speech. Calling `speak()` multiple time should queue the utterances (the caller is suspended until its utterance is played).
    * `AndroidTtsEngine` is an implementation using the native Android TTS
    * `PublicationSpeechSynthesizer` is the high-level orchestrator that an app can use to read a whole publication aloud.
* In the Test App:
    * Removed the legacy TTS implementation which was entirely in the test app
    * Refactored the `BaseReaderFragment` options menu to have a single one. Before EPUB had a specific options menu layout which contained the TTS button.
    * Added a `TtsViewModel` which will be composed inside `ReaderViewModel` to make it easier to look at TTS stuff.
    * Added a Jetpack Compose `overlay` view on top of the reader in `VisualReaderFragment`. This is used to display the TTS controls above the publication, but also to prevent user interaction on the pages while the TTS is running.

## Unfinished stuff

The following can be done in follow-up PRs as it's not critical right now:

* Supporting custom filters in the `HtmlResourceContentIterator` to skip specific HTML nodes.
* Better support for various HTML tags and ARIA attributes.
* Adding a `ContentService` implementation for PDF documents.

---

* Closes #68


readium/navigator-tts/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+    id("kotlin-android")
+    id("kotlin-parcelize")
+    id("maven-publish")
+    id("org.jetbrains.dokka")
+}
+
+android {
+    resourcePrefix = "readium_"
+
+    compileSdk = 31
```
stevenzeck, line 18: Should compile and target SDK 32

readium/navigator-tts/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+        targetSdk = 31
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
+    }
+    kotlinOptions {
+        jvmTarget = "1.8"
+        freeCompilerArgs = freeCompilerArgs + "-Xopt-in=kotlin.RequiresOptIn"
```
stevenzeck, line 31: Replace `-Xopt-in` with `-opt-in`

mickael-menu: Thanks, I'll fix your suggestions when I get back to Kotlin.

readium/navigator/src/main/java/org/readium/r2/navigator/Navigator.kt
```diff
@@ -104,11 +104,19 @@ interface NavigatorDelegate {
...
     /**
      * Current reading progression direction.
      */
     val readingProgression: ReadingProgression
 
+    /**
+     * Returns the [Locator] to the first content element that begins on the current screen.
+     */
+    @ExperimentalReadiumApi
+    suspend fun firstVisibleElementLocator(): Locator? =
```
mickael-menu, line 117: This was added to start the TTS from the currently visible page instead of the start of the publication. Using the `progression` in the `currentLocator` was not precise enough, and `firstVisibleElementLocator()` makes sure to target the first HTML element fully visible or starting on the page.

readium/shared/src/main/java/org/readium/r2/shared/OptIn.kt
```diff
@@ -25,6 +25,14 @@ annotation class InternalReadiumApi
 @Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.TYPEALIAS, AnnotationTarget.PROPERTY)
 annotation class ExperimentalReadiumApi
 
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING,
+    message = "This is a delicate API and its use requires care. Make sure you fully read and understand documentation of the declaration that is marked as a delicate API."
+)
+@Retention(value = AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.TYPEALIAS, AnnotationTarget.PROPERTY)
+annotation class DelicateReadiumApi
```
mickael-menu, line 34: Inspired by `@DelicateCoroutinesApi`. We can use it to tag a dangerous Readium API that should be used after carefully reading the doc.

readium/shared/src/main/java/org/readium/r2/shared/fetcher/Resource.kt
```diff
@@ -102,6 +104,15 @@ interface Resource : SuspendingCloseable {
     suspend fun readAsXml(): ResourceTry<ElementNode> =
         read().mapCatching { XmlParser().parse(ByteArrayInputStream(it)) }
 
+    /**
+     * Reads the full content as a [Bitmap].
+     */
+    suspend fun readAsBitmap(): ResourceTry<Bitmap> =
```
mickael-menu, line 110: Not necessary for this PR, but this seems to be a useful helper. I used it in the example of the `content.md` user guide to build an index of images.

readium/shared/src/main/java/org/readium/r2/shared/util/Language.kt
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright 2022 Readium Foundation. All rights reserved.
+ * Use of this source code is governed by the BSD-style license
+ * available in the top-level LICENSE file of the project.
+ */
+
+package org.readium.r2.shared.util
+
+import java.util.*
+
+class Language private constructor(val code: String, val locale: Locale) {
```
mickael-menu, line 11: I introduced this `Language` to have a higher language construct that doesn't depend on the JVM, and to add some useful helpers such as `removeRegion()`.

test-app/src/main/java/org/readium/r2/testapp/epub/UserSettings.kt
```diff
@@ -572,29 +572,6 @@ class UserSettings(var preferences: SharedPreferences, val context: Context, pri
                     }
                 })
 
-        // Speech speed
```
mickael-menu, line 575: I removed the legacy TTS implementation that was entirely in the test app.

test-app/src/main/java/org/readium/r2/testapp/reader/AudioReaderFragment.kt
```diff
@@ -37,7 +37,6 @@ import kotlin.time.ExperimentalTime
 @OptIn(ExperimentalMedia2::class, ExperimentalTime::class, ExperimentalCoroutinesApi::class)
 class AudioReaderFragment : BaseReaderFragment(), SeekBar.OnSeekBarChangeListener {
 
-    override val model: ReaderViewModel by activityViewModels()
```
mickael-menu, line 40: This could be moved to `BaseReaderFragment` to simplify the subclasses.

test-app/src/main/java/org/readium/r2/testapp/reader/EpubReaderFragment.kt
```diff
@@ -175,19 +151,20 @@ class EpubReaderFragment : VisualReaderFragment(), EpubNavigatorFragment.Listene
 
     override fun onCreateOptionsMenu(menu: Menu, menuInflater: MenuInflater) {
         super.onCreateOptionsMenu(menu, menuInflater)
-        menuInflater.inflate(R.menu.menu_epub, menu)
```
mickael-menu, line 178: All the options menu items are now in a single layout inflated by `BaseReaderFragment`.

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderActivity.kt
```diff
@@ -38,20 +39,20 @@ import org.readium.r2.testapp.outline.OutlineFragment
  */
 open class ReaderActivity : AppCompatActivity() {
 
-    private lateinit var modelFactory: ReaderViewModel.Factory
-    private lateinit var model: ReaderViewModel
+    private val model: ReaderViewModel by viewModels()
+
+    override fun getDefaultViewModelProviderFactory(): ViewModelProvider.Factory {
```
mickael-menu, line 44: This change seems pointless but I refactored it twice. At some point `TtsViewModel` was a proper Android `ViewModel` and so I had a composite view model factory.

test-app/src/main/java/org/readium/r2/testapp/reader/VisualReaderFragment.kt
```diff
@@ -51,49 +75,205 @@ abstract class VisualReaderFragment : BaseReaderFragment(), VisualNavigator.List
...
+
+        childFragmentManager.addOnBackStackChangedListener {
+            updateSystemUiVisibility()
+        }
+        binding.fragmentReaderContainer.setOnApplyWindowInsetsListener { container, insets ->
+            updateSystemUiPadding(container, insets)
+            insets
+        }
+
+        binding.overlay.setContent {
```
mickael-menu, line 98: I added a Jetpack Compose `overlay` view on top of the reader in `VisualReaderFragment`. This is used to display the TTS controls above the publication, but also to prevent user interaction on the pages while the TTS is running.

test-app/src/main/java/org/readium/r2/testapp/reader/VisualReaderFragment.kt
```diff
@@ -51,49 +75,205 @@ abstract class VisualReaderFragment : BaseReaderFragment(), VisualNavigator.List
...
+    private fun setupObservers() {
+        viewLifecycleOwner.lifecycleScope.launch {
+            viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                navigator.currentLocator
+                    .onEach { model.saveProgression(it) }
+                    .launchIn(this)
+
+                setupHighlights(this)
+                setupSearch(this)
+                setupTts(this)
```
mickael-menu, line 142: I split the setup to make it clearer to people wanting to integrate only one of these features.

test-app/src/main/java/org/readium/r2/testapp/shared/views/Group.kt
```diff
@@ -0,0 +1,26 @@
...
+import androidx.compose.material.ContentAlpha
+import androidx.compose.material.LocalContentAlpha
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+
+/**
+ * Sets the emphasis (alpha) of a group of [Composable] views.
+ */
+@Composable
+fun Group(lowEmphasis: Boolean = false, enabled: Boolean = true, content: @Composable () -> Unit) {
```
mickael-menu, line 18: This is not really necessary, but I find the [recommended approach for de-emphasizing elements](https://developer.android.com/jetpack/compose/compositionlocal) really heavy.

readium/navigator/src/main/java/org/readium/r2/navigator/tts/AndroidTtsEngine.kt
```diff
@@ -0,0 +1,274 @@
...
+        val error: EngineError =
+            EngineError.getOrDefault(code)
+    }
+
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
+
+    /**
+     * Utterances to be synthesized, in order of [speak] calls.
+     */
+    private val tasks = Channel<UtteranceTask>(Channel.BUFFERED)
```
mickael-menu, line 71: Using a `Channel` ensures that the utterances are played in the same order as the `speak()` calls.

readium/navigator/src/main/java/org/readium/r2/navigator/tts/PublicationSpeechSynthesizer.kt
```diff
@@ -0,0 +1,553 @@
...
+        }
+    }
+
+    /**
+     * Cancels the previous playback-related job and starts a new one with the given suspending
+     * [block].
+     *
+     * This is used to interrupt on-going commands.
+     */
+    private fun replacePlaybackJob(block: suspend CoroutineScope.() -> Unit) {
```
mickael-menu, line 527: That's the solution I came up with to be able to skip utterances or interrupt a running utterance. As only one playback job is allowed at any time, when you start a new command the previous one is cancelled.

test-app/src/main/java/org/readium/r2/testapp/reader/tts/TtsControls.kt
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * TTS controls bar displayed at the bottom of the screen when speaking a publication.
+ */
+@OptIn(ExperimentalReadiumApi::class)
+@Composable
+fun TtsControls(model: TtsViewModel, modifier: Modifier = Modifier) {
+    val showControls by model.state.asStateWhenStarted { it.showControls }
+    val isPlaying by model.state.asStateWhenStarted { it.isPlaying }
+    val settings by model.state.asStateWhenStarted { it.settings }
```
mickael-menu, line 36: I started with the usual:

```kotlin
val state by model.state.flowWithLocalLifecycle().collectAsState()
```

But as I used a single `State` object that is updated every time a word is spoken, that triggered a lot of recomposition. I came up with this `asStateWhenStarted` extension that select a property of the global state instead, and is basically (it's a bit more complicated, see `Flow.kt`) equivalent to:

```kotlin
val showControls by remember {
    model.state
        .map { it.showControls }
        .flowWithLocalLifecycle()
}.collectAsState()
```

PR #123, opened 2022-05-13
Fix caching with `TransformingResource`

mickael-menu:
#### Shared

* `TransformingResource` now caches its content by default, as it is the correct behavior in most cases. Set `cacheBytes = false` explicitly to revert to the previous behavior.

---

This fixes an issue with duplicated fetches when serving a packaged WebPub in the server.
See for more context the bug report: https://readium.slack.com/archives/C703MSTQU/p1652257069782699?thread_ts=1650958628.109609&cid=C703MSTQU

### DIFF
--- a/readium/shared/src/main/java/org/readium/r2/shared/fetcher/Resource.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/fetcher/Resource.kt
@@ -220,69 +220,17 @@ abstract class ProxyResource(protected val resource: Resource) : Resource {
 }
 
 /**
- * Caches the members of [resource] on first access, to optimize subsequent accesses.
- *
- * This can be useful when reading [resource] is expensive.
- *
- * Warning: bytes are read and cached entirely the first time, even if only a [range] is requested.
- * So this is not appropriate for large resources.
- */
-@Deprecated("If you were caching a TransformingResource, build it with cacheBytes set to true." +
-        "Otherwise, please report your use case.",
-    level = DeprecationLevel.ERROR)
-class CachingResource(private val resource: Resource) : Resource {
-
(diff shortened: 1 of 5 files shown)


readium/streamer/src/main/java/org/readium/r2/streamer/fetcher/HtmlInjector.kt
```diff
@@ -42,7 +42,7 @@ internal class HtmlInjector(
     private suspend fun inject(resource: Resource): Resource = object : TransformingResource(resource) {
 
         override suspend fun transform(data: ResourceTry<ByteArray>): ResourceTry<ByteArray> =
-            resource.read().mapCatching {
```
mickael-menu, line 45: Ups, we were reading the full resource instead of transforming the given data, which read the resource twice.

readium/shared/src/main/java/org/readium/r2/shared/fetcher/Resource.kt
```diff
@@ -219,70 +219,18 @@ abstract class ProxyResource(protected val resource: Resource) : Resource {
...
 
-/**
- * Caches the members of [resource] on first access, to optimize subsequent accesses.
- *
- * This can be useful when reading [resource] is expensive.
- *
- * Warning: bytes are read and cached entirely the first time, even if only a [range] is requested.
- * So this is not appropriate for large resources.
- */
-@Deprecated("If you were caching a TransformingResource, build it with cacheBytes set to true." +
```
mickael-menu, line 230: I think it's fine to remove this now. The `@Deprecated` was emitting an error so nobody should be using it by now.

qnga (approved):

PR #97, opened 2022-03-24
Move the Gradle plugins versions into `settings.gradle.kts`

mickael-menu:
While integrating the latest Kotlin toolkit using the Gradle Kotlin DSL as a Git submodule, I faced this error:
Screenshot 2022-03-24 at 16 41 58

According to [this Gradle forum post](https://discuss.gradle.org/t/error-plugin-already-on-the-classpath-must-not-include-a-version/31814/5), we can fix this by moving the version management to `settings.gradle.kts` instead.

Some additional info from the official documentation: https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_version_management

PR #126, opened 2022-06-02
Upgrade dependencies and Gradle

mickael-menu:
EDIT: The following is outdated, I forgot about this other issue:
 
* https://github.com/readium/kotlin-toolkit/pull/97

---

I moved the Gradle plugin versions from the [`settings.gradle.kts`'s `pluginManagement` section](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_version_management) to the [root `build.gradle.kts` file](https://developer.android.com/studio/releases/gradle-plugin#updating-plugin) instead. The [AGP Upgrade Assistant](https://developer.android.com/studio/build/agp-upgrade-assistant) was not working with `settings.gradle.kts` and having the versions in the root `build.gradle.kts` seems to be enough:

>You can specify the plugin version in either the File > Project Structure > Project menu in Android Studio, or the top-level build.gradle file. The plugin version applies to all modules built in that Android Studio project.
> https://developer.android.com/studio/releases/gradle-plugin#updating-plugin

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,8 +7,8 @@
 import org.jetbrains.dokka.gradle.DokkaTaskPartial
 
 plugins {
-    id("com.android.application") version "7.1.3" apply false
-    id("com.android.library") version "7.1.3" apply false
+    id("com.android.application") version "7.2.1" apply false
+    id("com.android.library") version "7.2.1" apply false
     id("org.jetbrains.kotlin.android") version "1.6.21" apply false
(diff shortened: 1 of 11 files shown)


mickael-menu:
I'm reverting the move of Gradle plugin versions back to `settings.gradle.kts`, I forgot about this other issue...

* https://github.com/readium/kotlin-toolkit/pull/97

qnga (approved):

PR #133, opened 2022-07-01
Fix comparison of `Decoration` objects

mickael-menu:
#### Navigator

* `Decoration.extras` is now a `Map<String, Any>` instead of `Bundle`. You will need to update your app if you were storing custom data in `extras`, for example:
    ```kotlin
    val decoration = Decoration(...,
        extras = mapOf("id" to id)
    )
    val id = decoration.extras["id"] as? Long
    ```

#### Navigator

* `Decoration` objects are now properly comparable with `equals()`.

---

Using `Bundle` for `Decoration.extras` broke the contract of data classes: `Decoration` couldn't be compared with `==` and was not fully immutable. Additionally, it added a strong dependency to Android.


test-app/src/main/java/org/readium/r2/testapp/reader/ReaderViewModel.kt
```diff
@@ -129,11 +129,11 @@ class ReaderViewModel(
             id = "$id-$idSuffix",
             locator = locator,
             style = style,
-            extras = Bundle().apply {
+            extras = mapOf(
                 // We store the highlight's database ID in the extras bundle, for easy retrieval
                 // later. You can store arbitrary information in the bundle.
```
qnga, line 134:
```suggestion
                // We store the highlight's database ID in the extras map, for easy retrieval
                // later. You can store arbitrary information in the map.
```

test-app/src/main/java/org/readium/r2/testapp/reader/VisualReaderFragment.kt
```diff
@@ -283,7 +283,7 @@ abstract class VisualReaderFragment : BaseReaderFragment(), VisualNavigator.List
             val decoration = event.decoration
             // We stored the highlight's database ID in the `Decoration.extras` bundle, for
             // easy retrieval. You can store arbitrary information in the bundle.
```
qnga, line 285:
```suggestion
            // We stored the highlight's database ID in the `Decoration.extras` map, for
            // easy retrieval. You can store arbitrary information in the map.
```

qnga (approved):

PR #137, opened 2022-07-08
Navigator: Fix `currentLocator` with FXL and prevent unnecessary refreshes 

mickael-menu:
#### Navigator

* Fixed emitting `currentLocator` with fixed layout EPUBs.
* Prevent refreshing an already loaded EPUB resource when jumping to a `Locator` in it.


readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
```diff
@@ -655,11 +648,14 @@ class EpubNavigatorFragment private constructor(
         get() = if (::resourcePager.isInitialized) resourcePager.adapter as? R2PagerAdapter
             else null
 
-    private val currentFragment: R2EpubPageFragment? get() =
```
mickael-menu, line 658: This property name was really ambiguous and caused the problem in the first place. It was `null` for an FXL resource.

PR #138, opened 2022-07-09
Update Android Gradle Plugin to 7.2.1

armendh:
upgrade com.android.application to 7.2.1
upgrade org.jetbrains.kotlin.android to 1.7.10

### DIFF
--- a/readium/navigator-media2/build.gradle.kts
+++ b/readium/navigator-media2/build.gradle.kts
@@ -20,7 +20,6 @@ android {
     defaultConfig {
         minSdk = 21
         targetSdk = 31
-        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     }
     compileOptions {
         sourceCompatibility = JavaVersion.VERSION_1_8
@@ -28,7 +27,10 @@ android {
     }
     kotlinOptions {
         jvmTarget = "1.8"
-        freeCompilerArgs = freeCompilerArgs + "-Xopt-in=kotlin.RequiresOptIn"
+        freeCompilerArgs = freeCompilerArgs + listOf(
+            "-opt-in=kotlin.RequiresOptIn",
+            "-Xjvm-default=all"
+        )
     }
     buildTypes {
         getByName("release") {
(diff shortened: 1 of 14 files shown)


readium/streamer/build.gradle.kts
```diff
@@ -13,11 +13,12 @@ plugins {
 }
 
 android {
+    resourcePrefix = "readium_"
```
stevenzeck, line 16: Should this be added in the other modules too (except for navigator-media2 which already has it)?

readium/navigator-media2/build.gradle.kts
```diff
@@ -20,15 +20,17 @@ android {
...
     compileOptions {
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
     kotlinOptions {
         jvmTarget = "1.8"
-        freeCompilerArgs = freeCompilerArgs + "-Xopt-in=kotlin.RequiresOptIn"
+        freeCompilerArgs = freeCompilerArgs + listOf(
+            "-opt-in=kotlin.RequiresOptIn",
+            "-Xjvm-default=all"
```
stevenzeck, line 32: Is there a specific need for `Xjvm-default`?

settings.gradle.kts
```diff
@@ -19,10 +19,10 @@ pluginManagement {
...
 
     plugins {
-        id("com.android.application") version ("7.1.0")
-        id("com.android.library") version ("7.1.0")
-        id("org.jetbrains.kotlin.android") version ("1.6.10")
-        id("org.jetbrains.dokka") version ("1.6.10")
+        id("com.android.application") version ("7.2.1")
+        id("com.android.library") version ("7.2.1")
+        id("org.jetbrains.kotlin.android") version ("1.7.10")
+        id("org.jetbrains.dokka") version "1.7.0"
```
stevenzeck, line 25: Can you add the parenthesis back around the version number here (for consistency)?

test-app/src/main/java/org/readium/r2/testapp/db/Database.kt
```diff
@@ -39,7 +39,7 @@ abstract class BookDatabase : RoomDatabase() {
                     context.applicationContext,
                     BookDatabase::class.java,
                     "books_database"
-                ).build()
+                ).allowMainThreadQueries().build()
```
stevenzeck, line 42: Same comment as removing `suspend`.

stevenzeck (changes requested):
Thanks @armendh!  I left a few comments.  This would be a good thing to merge either right before a Readium release or right after.

PR #143, opened 2022-07-29
EPUB: Fix page swipes while selecting text

### LINKED ISSUE #86: Moving to previous page when selecting text
When doing a selection where the end position x is > to the start position x, this is interpreted as a "swipe towards right" gesture, and it moves to the previous page. The selection pop-up menu allears on the new page, whereas the selection is where it should be, on the page that just disappeared. Tested in R2 testapp, develop, and a R2-based app. 
- Page with the pop-up (incorrect)
Screenshot_20220214-182601_Aldiko
- Page with the selection:
Screenshot_20220214-182647_Aldiko

mickael-menu:
#### Navigator

* [#86](https://github.com/readium/kotlin-toolkit/issues/86) Fixed page swipes while selecting text in an EPUB resource.
* The `onTap` event is not sent when an EPUB text selection is active anymore, to prevent showing the app bar while dismissing a selection.

---

* Fixes #86


qnga (approved):

aviashiet:
Thanks Mickael
Page swipe is fixed but still previous page selection is happening .
Can you suggest for this also that how to stop previous page text selection when selection at top of any page .

mickael-menu:
I don't know why this is happening and can't reproduce it on my devices, so I have no idea at this point.

aviashiet:
Thanks ,May I know on which device you have checked ?

mickael-menu:
Pixel 3a and 6a

aviashiet:
I observe that it works fine on large screen devices like Samsung tabs ,Pixel 3a etc.
on large screens pages are divided in large paragraph as compared  to small screen devices , I guess this problem happens only on first  small paragraph on top of every pages only.

mickael-menu:
I don't have smaller devices, but I tried on a Nexus One emulator (3.7") and couldn't reproduce the problem either.

mickael-menu:
Would you be able to debug the issue yourself as you can reproduce it on one of your devices?

mickael-menu:
I finally reproduced the issue, but the prospect is not good: https://github.com/readium/kotlin-toolkit/issues/325

PR #262, opened 2022-08-09
1/3 Replace legacy user settings API

mickael-menu:
## Review notes

The Settings API PRs are numbered in the order of merge, but should be reviewed in reverse order, starting with https://github.com/readium/kotlin-toolkit/pull/141.

However, you might want to start by reading the documentation that is in this PR, as it is the most up-to-date:

* [Configuring the Navigator](https://github.com/readium/kotlin-toolkit/blob/feature/settings+replace-legacy/docs/guides/navigator-settings.md)
* [Adding custom fonts to the EPUB navigator](https://github.com/readium/kotlin-toolkit/blob/feature/settings%2Breplace-legacy/docs/guides/epub-custom-fonts.md)
* [Migration Guide](https://github.com/readium/kotlin-toolkit/blob/feature/settings+replace-legacy/docs/migration-guide.md)

This PR:

* replaces the legacy user settings in the EPUB navigator with the new ones in a backward compatible way
* adds the migration guide


readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
```diff
@@ -271,62 +262,50 @@ class EpubNavigatorFragment private constructor(
...
-                        R2PagerAdapter(childFragmentManager, resourcesSingle)
-                    }
-                }
             }
         }
-        adapter.listener = PagerAdapterListener()
 
-        resourcePager.adapter = adapter
-        resourcePager.direction = publication.metadata.effectiveReadingProgression
+        resetAdapter()
```
mickael-menu, line 305: The adapter can now be reset when changing settings that require a relayout (e.g. reading progression).

readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
```diff
@@ -221,7 +216,9 @@ class EpubNavigatorFragment private constructor(
     private lateinit var resourcesSingle: List<PageResource>
     private lateinit var resourcesDouble: List<PageResource>
 
-    lateinit var preferences: SharedPreferences
```
mickael-menu, line 224: The `EpubNavigatorFragment` is in "legacy mode" (with `SharedPreferences` settings) if a `baseUrl` is provided. If it is `null`, we switch to the new Settings API.

readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
```diff
@@ -781,7 +793,7 @@ class EpubNavigatorFragment private constructor(
     }
 
     override val readingProgression: ReadingProgression
-        get() = publication.metadata.effectiveReadingProgression
+        get() = viewModel.readingProgression
```
mickael-menu, line 796: The `readingProgression` is now dynamic as it's also a setting.

readium/navigator/src/main/java/org/readium/r2/navigator/epub/css/ReadiumCss.kt
```diff
@@ -162,6 +161,9 @@ data class ReadiumCss(
             Layout.HtmlDir.Rtl -> "rtl"
         } ?: return
 
+        // Removes any dir attributes in html/body.
+        content.replace(0, content.length, content.replace(dirRegex, "$1"))
+
```
qnga, line 166: Are you sure about that? It goes against the EPUB spec, and doesn't match Readium CSS recommendations either, as far as I could see.

mickael-menu: Yeah because if an HTML tag has two `dir` attributes, the web view displays an error.

qnga: Sure, I meant that Readium CSS doesn't recommend to do anything when the direction is LTR. You're preventing a use case with RTL in some resources from working, without any benefit. I wonder if we shouldn't have a setting "Force language and direction in each resource".

mickael-menu: If we don't explicitly override any `dir` attribute, we can't set the reading progression of a book declared as RTL to LTR. Besides, the reader would break if not all resources use the same `dir`, as the global reading progression implies that it flows in a single direction across resources.

> You're preventing a use case with RTL in some resources from working, without any benefit.

Which use case?

> I wonder if we shouldn't have a setting "Force language and direction in each resource".

The `readingProgression` and `language` settings are just that. When you don't want to force it, you reset them to `null`.

qnga: > Besides, the reader would break if not all resources use the same dir, as the global reading progression implies that it flows in a single direction across resources.

Good point, EPUB might be a bit too optimistic on this.

> The readingProgression and language settings are just that. When you don't want to force it, you reset them to null.

As I understand it, `null` means automatic behavior depending on publication metadata, it does not mean "no action".

mickael-menu: Multi-reading progression publications are supported by EPUB but we decided to ignore it in Readium as it breaks affordances across resources. Which means we need to override the `dir` attributes in all cases.

qnga (approved):
It's hard to say that I saw and thought about everything, but let's go. I might have more insight after working on PDF settings.

PR #264, opened 2022-08-21
Target SDK 33 and use Material 3

stevenzeck:
This PR updates all of the modules to compile to and target Android SDK 33 (Android 13).  In addition, the theme has been updated to use Material 3.

- In each module's `build.gradle.kts`, the `compileSdk` and `targetSdk` have been changed from 32 to 33 to coincide with Android 13
- In doing the last point, R2FXLLayout decided the `GestureListener` in that file could not be smart cast to `OnScaleGestureListener` because it's a mutable property.  Probably some change in `ScaleGestureDetector` in the SDK between 32 and 33, so just asserted it's not null
- Also in changing the SDK to 33, Android Studio/Gradle strongly suggested upgrading the Android Gradle Plugin, which I did (7.2.2 to 7.2.3)
- `MenuItem` is no longer nullable when implementing `OnActionExpandListener`, so made that change
- Removed the hard-coded colors in the audiobook player
- Changed the theme from `Theme.MaterialComponents.Light.DarkActionBar` to `Theme.Material3.Light`. With `DynamicColors.applyToActivitiesIfAvailable(this)` in the `Application`, a device running Android 12 or higher will use the device's color theme.  Android 11 and below will have the default color theme, which is purplish


mickael-menu (changes requested):
Thanks Steven, looking good!

I just have two comments:

Could we use the default baseline color scheme instead of generating a custom one based off `#73b02c`? To simplify the test app config files.

The background of the media player is a bit intense 😄 I would use a neutral background color for it.

Screenshot_1661505970

stevenzeck:
Done.  I updated the original PR comment to reflect the last commit.

mickael-menu (approved):

PR #269, opened 2022-09-08
Add accessibility metadata

qnga:


readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/MetadataAdapter.kt
```diff
@@ -0,0 +1,465 @@
...
+        var remainingItems: List<MetadataItem> = globalItems
+
+        val coverId: String? = remainingItems
+            .takeFirstWithProperty("cover")
+            .let { remainingItems = it.second; it.first }
+            ?.value
+
+        val duration: Double? = DurationAdapter()
+            .adapt(remainingItems)
+            .let { remainingItems = it.second; it.first }
```
mickael-menu, line 48: I don't find this pattern very readable. Maybe we can improve it?

We could modify in place a `MutableList` but if you prefer keeping the immutability:

There are a quite a lot of adapters following this template:

```kotlin
fun adapt(items: List<MetadataItem>): Pair<XXX, List<MetadataItem>>
```

Let's say we add a `MetadataItemAdapter` function interface:

```kotlin
data class MetadataItemResult<T> {
    val result: T
    val remaining: List<MetadataItem>
}

fun interface MetadataItemAdapter<T> {
    fun adapt(items: List<MetadataItem>): MetadataItemResult<T>
}
```

Then we should be able to add an inline `extract` helper:

```kotlin
var remainingItems: List<MetadataItem> = globalItems

fun <T> extract(adapter: MetadataItemAdapter<T>): T =
    adapter.adapt(remainingItems)
        .let { remainingItems = it.remaining; it.result }

val duration: Double? = extract(DurationAdapter())
val languages: List<String> = extract(LanguageAdapter(readingProgression))
```

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/AccessibilityAdapter.kt
```diff
@@ -0,0 +1,168 @@
...
+        isWCAG_20_AA(value) -> Accessibility.Profile.EPUB_A11Y_10_WCAG_20_AA
+        isWCAG_20_AAA(value) -> Accessibility.Profile.EPUB_A11Y_WCAG_20_AAA
+        else -> null
+    }
+
+    private fun isWCAG_20_A(value: String) = value in setOf(
+        "EPUB Accessibility 1.1 - WCAG 2.0 Level A",
+        "http://idpf.org/epub/a11y/accessibility-20170105.html#wcag-a",
+        "http://www.idpf.org/epub/a11y/accessibility-20170105.html#wcag-a",
+        "http://wwwidpf.org/epub/a11y/accessibility-20170105.html#wcag-a",
```
mickael-menu, line 149: Where is this one coming from?

mickael-menu (changes requested):

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/AccessibilityAdapter.kt
qnga: That's a weird typo. I meant `https://idpf.org/epub/a11y/accessibility-20170105.html#wcag-a`.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/ResourceAdapter.kt
```diff
@@ -1,107 +1,39 @@
...
+    private val itemrefByIdref = spine.itemrefs
+        .associateBy(Itemref::idref)
 
-    fun create(): Manifest {
-        // Compute metadata
-        val metadata = pubMetadata.metadata()
-        val metadataLinks = links.mapNotNull(::mapEpubLink)
-
-        // Compute links
+    fun adapt(): Pair<List<Link>, List<Link>> {
```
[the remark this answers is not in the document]
qnga: Why not

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/MetadataAdapter.kt
qnga: Yeah, I was not very happy with this pattern. I couldn't find a better one though.
If I understand correctly your proposal, we would have to define `extract` in every place where a `remainingItems` list is created.

Mutability in a very small and specialized class might be reasonable. 

```kotlin
data class MetadataItemsHolder(var items: List<MetadataItem>) {
    fun <T> adapt(adapter: MetadataAdapter<T>): T {
        adapter.adapt(items)
              .let { items = it.remaining; it.result }
    }
}
```

mickael-menu: > If I understand correctly your proposal, we would have to define extract in every place where a remainingItems list is created.

That's right.

> Mutability in a very small and specialized class might be reasonable.

I agree, that looks good.

mickael-menu (approved):

PR #271, opened 2022-09-16
Fix `xml:lang` injection when blank

mickael-menu:
When a resource contains an empty `lang=""` tag in `<body>` or `<html>`, an extra `xml:lang` attribute was added which broke the WebView rendering.

See https://readium.slack.com/archives/C703MSTQU/p1663081573775099


mrifni:
i have tested this PR and its working now. thanks

PR #279, opened 2022-09-28
Fix bottom tabs and publication detail in testapp

stevenzeck:
This PR fixes two relatively small issues with the testapp.

1. When a publication doesn't have a cover image (in the `images` array in the OPDS feed), the publication detail screen crashes.  Added a check to make sure it has an image before trying to load it
2. Under the scenario outlined below, the bottom navigation selected tab would break.  This was reproduced in a minimal repo.  The fix was to add separate nav graphs for each of the tabs, in addition to the main one.  Borrowed from https://github.com/android/architecture-components-samples/tree/master/NavigationAdvancedSample/app/src/main/res/navigation

Selected Tab Issue

1. Select the Catalogs tab
2. Select a feed to go to the catalog screen
3. Select the Bookshelf tab
4. Select the Catalogs tab

Result:  It navigates to the catalog screen from step 2, however the Bookshelf tab is still selected.

### DIFF
--- a/test-app/src/main/java/org/readium/r2/testapp/catalogs/PublicationDetailFragment.kt
+++ b/test-app/src/main/java/org/readium/r2/testapp/catalogs/PublicationDetailFragment.kt
@@ -46,8 +46,10 @@ class PublicationDetailFragment : Fragment() {
         super.onViewCreated(view, savedInstanceState)
         (activity as MainActivity).supportActionBar?.title = publication?.metadata?.title
 
-        Picasso.get().load(publication?.images?.first()?.href)
-            .into(binding.catalogDetailCoverImage)
+        if (publication?.images?.isNotEmpty() == true) {
+            Picasso.get().load(publication?.images?.first()?.href)
+                .into(binding.catalogDetailCoverImage)
+        }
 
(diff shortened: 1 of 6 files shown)


test-app/src/main/java/org/readium/r2/testapp/catalogs/PublicationDetailFragment.kt
```diff
@@ -46,8 +46,10 @@ class PublicationDetailFragment : Fragment() {
         super.onViewCreated(view, savedInstanceState)
         (activity as MainActivity).supportActionBar?.title = publication?.metadata?.title
 
-        Picasso.get().load(publication?.images?.first()?.href)
-            .into(binding.catalogDetailCoverImage)
+        if (publication?.images?.isNotEmpty() == true) {
+            Picasso.get().load(publication?.images?.first()?.href)
+                .into(binding.catalogDetailCoverImage)
+        }
```
qnga, line 52: Cosmetic suggestion

```suggestion
        publication?.images?.firstOrNull()
            ?.let { Picasso.get().load(it.href) }
            ?.into(binding.catalogDetailCoverImage)
```

mickael-menu (approved):
Thank you for tackling these issues 👍

PR #283, opened 2022-10-05
Move dependency management to version catalogs

stevenzeck:
This PR moves a majority of dependency versioning from the `build.gradle.kts` files and into version catalogs.  There are a few dependencies in `shared` and `navigator` I left as is because they are only used in those modules, and they're hopefully going away soon anyways.

Android Studio also recommended migrating the package names from `AndroidManifest.xml` and into `build.gradle.kts` namespace, and that is included here too.

I also changed the Kotlin plugins to use `kotlin()` instead of `id()`.

Finally, I removed the `jcenter()` methods in the settings gradle file and replaced with `maven(url = "https://jcenter.bintray.com")` to get rid of the deprecation.


mickael-menu (approved):
Thanks @stevenzeck!

I tested with the new submodule integration and the TOML catalog works fine.

PR #285, opened 2022-10-11
Fix integration as Git submodule

mickael-menu:
Recommend to use a [Gradle Composite build](https://docs.gradle.org/current/userguide/composite_builds.html) to integrate the Readium toolkit as a local clone. This simplifies significantly the setup in the host app.

I decided to use the same group and artifact IDs as JitPack. This allows an app to write:

```
implementation 'com.github.readium.kotlin-toolkit:readium-shared'
```

to reference the official JitPack binaries. Then, just by adding `includeBuild 'readium'`, it will switch to a local submodule.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,6 +13,10 @@ plugins {
     id("org.jetbrains.dokka") apply true
 }
 
+allprojects {
+    group = "com.github.readium.kotlin-toolkit"
+}
+
 subprojects {
     tasks.register<Jar>("javadocsJar") {
         archiveClassifier.set("javadoc")
--- a/readium/navigator/src/main/java/org/readium/r2/navigator/media/ExoMediaPlayer.kt
+++ b/readium/navigator/src/main/java/org/readium/r2/navigator/media/ExoMediaPlayer.kt
@@ -114,7 +114,7 @@ class ExoMediaPlayer(
         .apply {
             setMediaSessionToken(mediaSession.sessionToken)
             setPlayer(player)
-            setSmallIcon(R.drawable.exo_notification_small_icon)
+            setSmallIcon(com.google.android.exoplayer2.ui.R.drawable.exo_notification_small_icon)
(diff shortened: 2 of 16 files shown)


settings.gradle.kts
```diff
@@ -45,15 +45,44 @@ dependencyResolutionManagement {
 rootProject.name = "Readium"
 
 include(":readium:adapters:pdfium:pdfium-document")
+project(":readium:adapters:pdfium:pdfium-document")
```
mickael-menu, line 48: By default Gradle will use the folder name as artifact ID, which is not suitable because we have several folders named "navigator", for example. "shared" might also conflit with other modules in a host project.

readium/navigator/src/main/java/org/readium/r2/navigator/media/ExoMediaPlayer.kt
```diff
@@ -114,7 +114,7 @@ class ExoMediaPlayer(
         .apply {
             setMediaSessionToken(mediaSession.sessionToken)
             setPlayer(player)
-            setSmallIcon(R.drawable.exo_notification_small_icon)
+            setSmallIcon(com.google.android.exoplayer2.ui.R.drawable.exo_notification_small_icon)
```
mickael-menu, line 117: We should not use unqualified `R` when referencing resources of nested dependencies, as it prevent using `android.nonTransitiveRClass=true`.

`android.nonTransitiveRClass=true` is enabled by default in Android Studio's project template and improves build performance.

build.gradle.kts
```diff
@@ -13,6 +13,10 @@ plugins {
     id("org.jetbrains.dokka") apply true
 }
 
+allprojects {
+    group = "com.github.readium.kotlin-toolkit"
```
mickael-menu, line 17: This allows to use the same dependency names as JitPack ones when using `includeBuild`. I also renamed the sub-projects to reflect the artifact IDs used with JitPack.

PR #293, opened 2022-11-03
Rewrite Settings API and introduce PDF settings

qnga:


qnga:
Not sure why lint fails. It's working locally.

mickael-menu (approved):

PR #314, opened 2022-12-23
Remove dependency on JitPack

mickael-menu:
JitPack is becoming more and more unreliable. We will move to Maven Central eventually for distributing Readium, but this PR addresses the other part of the problem by removing any dependency on JitPack in the project.

We still have two dependencies distributed through JitPack:

* `nanohttpd` for the HTTP server
* ChrisBanes's PhotoView for the CBZ navigator

As both of these will soon be deprecated and removed from the project, for now I just added the JAR files in the project.


qnga (approved):
Great idea!

PR #330, opened 2023-01-27
Emit `onJumpToLocator` when swiping across resources in scroll mode

mickael-menu:
#### Navigator

* Scroll mode: jumping between two EPUB resources with a horizontal swipe triggers the `Navigator.Listener.onJumpToLocator()` callback.
    * This can be used to allow the user to go back to their previous location if they swiped across chapters by mistake.


readium/navigator/src/main/java/org/readium/r2/navigator/R2BasicWebView.kt
```diff
@@ -209,22 +226,29 @@ open class R2BasicWebView(context: Context, attrs: AttributeSet) : WebView(conte
...
-                goLeft()
-            } else {
-                runJavaScript("readium.scrollLeft();") { success ->
-                    if (!success.toBoolean()) {
-                        goLeft()
+            when {
+                scrollMode ->
+                    goLeft(jump = true)
+
+                !this@R2BasicWebView.canScrollHorizontally(1) ->
```
qnga, line 243: Why did you change -1 to 1?

mickael-menu: That's a botched copy-paste.

qnga (approved):

PR #331, opened 2023-01-27
Adjust `EpubNavigatorFragment.firstVisibleElementLocator()` strategy

mickael-menu:
#### Changed

* `EpubNavigatorFragment.firstVisibleElementLocator()` now returns the first *block* element that is visible on the screen, even if it starts on previous pages.
    * This is used to make sure the user will not miss any context when restoring a TTS session in the middle of a resource.


qnga (approved):
Looks good.

PR #337, opened 2023-02-12
Add support for full remote publications

qnga:
- Refactor publication assets and parser interface.
- Introduce remote assets, including ZIP-packaged ones, and the ability to open a LCP-protected publication through its LCPL file.
- Status from LSD takes precedence over the old license after a failed license update.


readium/shared/src/main/java/org/readium/r2/shared/publication/asset/RemoteAsset.kt
```diff
@@ -0,0 +1,77 @@
...
+     * Creates a [RemoteAsset] from a [URL] and an optional media type hint.
+     *
+     * Providing a media type hint will improve performances when sniffing the media type.
+     */
+    constructor(url: URL, mediaTypeHint: String?) :
+        this(url, knownMediaType = null, mediaTypeHint = mediaTypeHint)
+
+    override suspend fun mediaType(): MediaType {
+        if (!::_mediaType.isInitialized) {
+            val bytes = { url.openConnection().getInputStream().use { it.readBytes() } }
```
mickael-menu, line 44: There's a [`URL.readBytes()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.net.-u-r-l/read-bytes.html) extension in the Kotlin stdlib which basically does the same thing.

readium/lcp/src/main/java/org/readium/r2/lcp/LcpService.kt
```diff
@@ -89,6 +111,11 @@ interface LcpService {
     fun contentProtection(authentication: LcpAuthenticating = LcpDialogAuthentication()): ContentProtection =
         LcpContentProtection(this, authentication)
 
+    /**
+     * Builds a [PublicationAsset] to open a LCP-protected publication from its license file.
+     */
+    fun remoteAssetForLicense(license: File): Try<PublicationAsset, LcpException>
```
mickael-menu, line 117: The license might not be stored on the file system, so I think we need a variant with maybe `ByteArray`?

readium/lcp/src/main/java/org/readium/r2/lcp/LcpDecryptor.kt
```diff
@@ -96,6 +104,9 @@ internal class LcpDecryptor(val license: LcpLicense?) {
                     }
             }
 
+            Timber.d("length ${_length.exceptionOrNull()} ${_length.getOrNull()}")
+            _length.onFailure { Timber.e(it) }
+
```
mickael-menu, line 109:
```suggestion
```

readium/lcp/src/main/java/org/readium/r2/lcp/LcpService.kt
qnga: Not for now because it couldn't be updated through a ByteArray.

readium/lcp/src/main/java/org/readium/r2/lcp/LcpDecryptor.kt
```diff
@@ -69,7 +69,14 @@ internal class LcpDecryptor(val license: LcpLicense?) {
         private val license: LcpLicense
     ) : Resource {
 
-        lateinit var _length: ResourceTry<Long>
+        private class Cache(
+            var startIndex: Int? = null,
+            val data: ByteArray = ByteArray(3 * AES_BLOCK_SIZE)
```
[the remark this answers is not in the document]
qnga: The `LcpDecryptor` needs to look around the data strictly matching the content to decipher. That means that in case of contiguous requests to the fetcher, data fetched by the decryptor are not contiguous. Every request starts slightly before the end of the previous one. This is an issue with remote publications because we have to make a new HTTP request every time instead of reusing the previous one.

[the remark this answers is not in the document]
mickael-menu: I see, thanks.

mickael-menu (approved):

Issue #36, opened 2021-08-15
3.0.0


mickael-menu:
:loudspeaker: We're beginning a new development cycle with breaking changes to prepare the 3.0 release of the Readium Kotlin Toolkit. Starting from now:
* `develop` will contain only critical bugfixes and dependency upgrades
* a new `v3` branch will host new features and breaking changes, until it's ready for production

Here's an (incomplete) list of changes to be expected.

## Major new features

* New TTS and audio navigators based on Jetpack media3, supporting background playback.
* Streaming of ZIP archives (e.g. LCP-protected audiobooks hosted on a remote server).

## Breaking changes

* [x] [Prefix all resources with `readium_`](https://github.com/readium/kotlin-toolkit/blob/540c80fc1c57f04d09db45764f103724177aedf9/readium/navigator/build.gradle.kts#L17).
* [x] Enable [Explicit API mode](https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md).
* [x] Remove localized exceptions strings from our modules. This is the responsibility of the apps.

### Shared

* [x] Remove the legacy [`compressedLength` `Link` property](https://github.com/readium/r2-shared-kotlin/blob/0d1416003f275b0d2757f0fcdc42840c7064758e/r2-shared/src/main/java/org/readium/r2/shared/fetcher/ArchiveFetcher.kt#L73).
* [x] Remove `Ref<Publication>` from the `Publication.Context` ([see this comment](https://github.com/readium/kotlin-toolkit/pull/129#discussion_r901111623)).
* [x] Remove `ReadingProgression.TTB/BTT/AUTO`.
* [x] Remove the `Parcelable` conformance on `Publication`, and deprecate the `PublicationRepository`. ❓ https://github.com/readium/kotlin-toolkit/issues/65

### Navigator

* [x] Remove all the `Activity` navigators.
* [x] Extract the `VisualNavigator`'s `presentation` and `goLeft`/`goRight` in a new interface dedicated to "flowing" publications, as it wouldn't make sense for a video-based publication.
* [x] Remove older settings API.
* [x] Remove the HTTP server.
* [x] Remove `highlight.js`

### LCP

* [x] We need to make the database calls in `License.kt` asynchronous and remove `allowMainThreadQueries()` on the Room setup (see https://github.com/readium/r2-lcp-kotlin/pull/116).
* [x] Fix lifecycle issue in `LcpDialogAuthenticating`

## Delayed until a next version
[...]

PR #343, opened 2023-03-28
Fixing R2EpubActivity to be usable without a server.

jalbertcory:
With the 2.3.0 update, the docs state that you do not need to provide a baseUrl, but this code was still validating it and throwing an exception.

This PR is removing the requirement for a baseUrl in R2EpubActivity.kt as the underlying EpubNavigatorViewModel.createFactory method is checking that case and serving the file.

I have this running in my app now without it erroring out, so the underlying server is working when null is passed causing it to be created.

### DIFF
--- a/readium/navigator/src/main/java/org/readium/r2/navigator/epub/R2EpubActivity.kt
+++ b/readium/navigator/src/main/java/org/readium/r2/navigator/epub/R2EpubActivity.kt
@@ -51,7 +51,7 @@ open class R2EpubActivity : AppCompatActivity(), IR2Activity, IR2Selectable, IR2
     override lateinit var publication: Publication
     override lateinit var publicationIdentifier: String
     override var bookId: Long = -1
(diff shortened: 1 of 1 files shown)


mickael-menu:
Thank you for opening this PR @jalbertcory. Unfortunately we won't merge it because `R2EpubActivity` will be removed in the next release (3.0).

See this issue for more details: https://github.com/readium/kotlin-toolkit/issues/36

If you want to use the serverless implementation, you need to use `EpubNavigatorFragment` directly in your app. I would recommend taking the time to create your own reader `Activity` (there's an example in the test app), but you can also copy and paste the `R2EpubActivity` and [its XML layout](https://github.com/readium/kotlin-toolkit/blob/830812b02a0597b4b22d0d37ec35846db24cb0a4/readium/navigator/src/main/res/layout/activity_r2_epub.xml) if you want a quick fix.

PR #353, opened 2023-05-25
Abstract the toolkit over publication source

qnga:
This is a major refactoring of `MediaType` and `Streamer`, introducing the new `Resource` and `Asset` types along with `AssetRetriever`, `MediaTypeRetriever` and `ContentProtectionSchemeRetriever`.

To be done in another PR:
- make error handling more consistent
- partially cache and buffer data for single file publications
- relax conditions for reading LCP-protected publications from shared storage
- user guide for the components replacing `Streamer`


readium/shared/src/main/java/org/readium/r2/shared/util/pdf/PdfDocument.kt
```diff
@@ -33,7 +33,7 @@ interface PdfDocumentFactory<T : PdfDocument> {
     suspend fun open(file: File, password: String?): T
 
     /** Opens a PDF from a Fetcher resource. */
-    suspend fun open(resource: Resource, password: String?): T
+    suspend fun open(resource: Fetcher.Resource, password: String?): T
```
mickael-menu, line 36: It feels to me that the `PdfDocument` should depend on the core `Resource` instead of the `Fetcher` one. I looked into the implementation and there's no use of the `link` that `Fetcher.Resource` adds, except in `CachingPdfDocumentFactory`. This one could use `name()` as an alternative, and not cache it if the `name()` is missing. Alternatively, we could have `open(name: String, resource: Resource)`.

readium/shared/src/main/java/org/readium/r2/shared/resource/Container.kt
```diff
@@ -0,0 +1,56 @@
...
+        /**
+         * Absolute path to the entry in the archive.
+         * It MUST start with /.
+         */
+        val path: String
+
+        /**
+         * Gets the entry name if any.
+         */
+        override suspend fun name(): ResourceTry<String?>
```
mickael-menu, line 31: What is being overridden here? It looks like the same signature as in `Resource`.

And what is the difference with `path` in the context of a `Container.Entry`?

readium/shared/src/main/java/org/readium/r2/shared/util/http/HttpResourceFactory.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.readium.r2.shared.resource.Resource
+import org.readium.r2.shared.resource.ResourceFactory
+import org.readium.r2.shared.util.Url
+
+class HttpResourceFactory(
+    private val httpClient: HttpClient
+) : ResourceFactory {
+
+    override suspend fun create(url: Url): Try<Resource, ResourceFactory.Error> {
+        if (!url.scheme.startsWith("http")) {
```
mickael-menu, line 19: Could be nice to add a helper similar to `url.isFile`:

```kotlin
fun Url.isHttp(): Boolean =
    scheme == "http" || scheme == "https"
```

test-app/src/main/res/values/strings.xml
```diff
@@ -169,10 +169,32 @@
...
+
+    <string name="r2.shared.resource.exception.bad_request">Invalid request which can\'t be processed</string>
+    <string name="r2.shared.resource.exception.not_found">Resource not found</string>
+    <string name="r2.shared.resource.exception.forbidden">You are not allowed to access the resource</string>
+    <string name="r2.shared.resource.exception.unavailable">The resource is currently unavailable, please try again later</string>
+    <string name="r2_shared_resource_exception_offline">The Internet connection appears to be offline</string>
+    <string name="r2.shared.resource.exception.out_of_memory">The resource is too large to be read on this device</string>
+    <string name="r2.shared.resource.exception.unexpected">An expected error occurred.</string>
+    <string name="r2.shared.resource.exception.other">A service error occurred</string>
+
```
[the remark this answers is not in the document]
qnga: I can't remember the reason why those strings have ended up here... I have not refactored `Resource.Exception` yet. I'll drop them for now.

test-app/src/main/java/org/readium/r2/testapp/utils/extensions/Uri.kt
```diff
@@ -10,15 +10,16 @@ import android.content.Context
...
-import org.readium.r2.shared.util.mediatype.MediaType
+import org.readium.r2.shared.error.Try
 import org.readium.r2.testapp.utils.ContentResolverUtil
 
 suspend fun Uri.copyToTempFile(context: Context, dir: File): Try<File, Exception> =
     try {
         val filename = UUID.randomUUID().toString()
-        val mediaType = MediaType.ofUri(this, context.contentResolver)
-        val file = File(dir, "$filename.${mediaType?.fileExtension ?: "tmp"}")
+        val extension = path
```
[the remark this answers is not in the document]
qnga: I can't remember. Is there any case when the file extension is not suitable?

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
```diff
@@ -6,49 +6,64 @@
...
+        get() = uri.path!!
 
-    val file: String
-        get() = url.file
+    val filename: String
+        get() = File(path).name
 
-    val extension: String?
-        get() = url.extension
+    val extension: String
```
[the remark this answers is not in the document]
qnga: That's not usual, probably because of the Java tradition, but that makes sense.

readium/streamer/src/main/java/org/readium/r2/streamer/extensions/Fetcher.kt
```diff
@@ -9,51 +9,18 @@
...
-
-/** Returns the resource data at the given [href], or throws a [Resource.Exception] */
-@Throws(Resource.Exception::class)
-internal suspend fun Fetcher.readBytes(href: String): ByteArray =
-    get(href).use { it.read().getOrThrow() }
-
 /** Returns the resource data as an XML Document at the given [href], or null. */
 internal suspend fun Fetcher.readAsXmlOrNull(href: String): ElementNode? =
-    get(href).use { it.readAsXml().getOrNull() }
+    get(href).use { it.readAsXml().successOrNull() }
```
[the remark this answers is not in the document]
qnga: I'm not sure what you mean. This `Fetcher` is still in use, though the `Fetcher` from the streamer module is deprecated.

readium/shared/src/main/java/org/readium/r2/shared/resource/Container.kt
qnga: I'm pretty sure I overrode `name` to get rid of complaints from the compiler about some ambiguity. I can't reproduce that though, it might be outdated.
`path` is a complete file path which can reflect some hierarchical organisation. `name` is just a name.

test-app/src/main/java/org/readium/r2/testapp/Readium.kt
```diff
@@ -11,41 +11,84 @@ import org.readium.adapters.pdfium.document.PdfiumDocumentFactory
...
-import org.readium.r2.streamer.Streamer
+import org.readium.r2.shared.util.http.HttpResourceFactory
+import org.readium.r2.shared.util.mediatype.MediaType
+import org.readium.r2.shared.util.mediatype.MediaTypeRetriever
+import org.readium.r2.streamer.PublicationFactory
 
 /**
  * Holds the shared Readium objects and services used by the app.
  */
 class Readium(context: Context) {
```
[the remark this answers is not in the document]
qnga: I tried something like this and that was not very compelling. Each component is useful and none can truly be hidden. I'm afraid we would end up with one-to-one mapping functions. For instance, the `AssetRetriever` must be used independently to download publications from LCP licenses.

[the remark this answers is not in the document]
mickael-menu: Alright, thank you.

test-app/src/main/java/org/readium/r2/testapp/utils/extensions/Uri.kt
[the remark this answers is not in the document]
mickael-menu: I noticed on some HTTP server that we don't get any extension, and we can't guarantee the file is properly named when importing a publication from the local storage.

That being said, Readium can work with a missing extension, so maybe we can keep things simple if it works like that.

readium/streamer/src/main/java/org/readium/r2/streamer/extensions/Fetcher.kt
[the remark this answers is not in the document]
mickael-menu: Yes, I thought these extensions were for the deprecated one, I didn't check the imports.

test-app/src/main/java/org/readium/r2/testapp/utils/extensions/Uri.kt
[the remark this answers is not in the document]
qnga: Yes, this piece of code is not used with HTTP and trusting the local storage sounds reasonable enough to me.

readium/shared/src/main/java/org/readium/r2/shared/publication/protection/ProtectionRetriever.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package org.readium.r2.shared.publication.protection
+
+import org.readium.r2.shared.asset.Asset
+import org.readium.r2.shared.util.mediatype.MediaTypeRetriever
+
+/**
+ * Retrieves [ContentProtection] schemes of assets.
+ */
+class ProtectionRetriever(
```
mickael-menu, line 15: How about `ContentProtectionRetriever` or `ContentProtectionSchemeRetriever`? For consistency with the other `Retriever` and all the things related to the DRM are named `ContentProtection`.

readium/shared/src/main/java/org/readium/r2/shared/publication/protection/ContentProtection.kt
```diff
@@ -7,56 +7,65 @@
...
      *
-     * @return A [ProtectedAsset] in case of success, null if the asset is not protected by this
+     * @return A [Asset] in case of success, null if the asset is not protected by this
      * technology or a [Publication.OpeningException] if the asset can't be successfully opened,
      * even in restricted mode.
      */
     suspend fun open(
-        asset: PublicationAsset,
+        asset: org.readium.r2.shared.asset.Asset,
+        drmScheme: String,
```
mickael-menu, line 49: How about using `drmScheme: Scheme` instead for better type safety?

Also, I'm starting to think that we should have just `value class Scheme(val uri: String)` instead of having a localized name with it. It will be easier to get a proper instance of `Scheme` from a raw URI, and we can let the app decide how to represent the name of a DRM.

readium/shared/src/main/java/org/readium/r2/shared/error/Error.kt
```diff
@@ -0,0 +1,47 @@
...
+        throwable.message ?: "Exception"
+
+    override val cause: Error? =
+        null
+}
+
+/**
+ * A basic [Error] implementation.
+ */
+class SimpleError(
```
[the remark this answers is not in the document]
qnga: Mmm, that would be the case if I remove the `cause` argument. That seems to work.

readium/shared/src/main/java/org/readium/r2/shared/publication/protection/ContentProtection.kt
qnga: There is  the same issue with restoring `MediaType`s from strings, though that's fine in practice. `name` is useless, but `fileExtension` is not.

readium/shared/src/main/java/org/readium/r2/shared/util/mediatype/MediaTypeRetriever.kt
```diff
@@ -0,0 +1,246 @@
...
+        bytes: () -> ByteArray,
+        mediaType: String? = null,
+        fileExtension: String? = null,
+    ): MediaType? {
+        return retrieve(bytes, mediaTypes = listOfNotNull(mediaType), fileExtensions = listOfNotNull(fileExtension))
+    }
+
+    /**
+     * Resolves a media type from bytes, e.g. from an HTTP response.
+     */
```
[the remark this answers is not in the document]
qnga: I don't think I've changed that. HTTP is just an example of case where you may want to pass bytes to the `MediaTypeRetriever`.

readium/shared/src/main/java/org/readium/r2/shared/util/mediatype/Extensions.kt
```diff
@@ -35,10 +37,19 @@ suspend fun HttpURLConnection.sniffMediaType(
 
     // TODO: The suggested filename extension, part of the HTTP header `Content-Disposition`.
 
+    val mediaTypeRetriever = MediaTypeRetriever(sniffers = sniffers)
```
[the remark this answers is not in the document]
qnga: Good idea. I implemented it as an extension in the `http` package: that's not so self-discovery, but opens the door to other HTTP-related extensions.

readium/shared/src/main/java/org/readium/r2/shared/util/mediatype/MediaTypeRetriever.kt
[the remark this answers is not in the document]
mickael-menu: Right, I read too fast and missed the `.e.g.`.

readium/shared/src/main/java/org/readium/r2/shared/resource/Resource.kt
```diff
@@ -0,0 +1,188 @@
...
+     * Reads the full content as an XML document.
+     */
+    suspend fun readAsXml(): ResourceTry<ElementNode> =
+        read().mapCatching { XmlParser().parse(ByteArrayInputStream(it)) }
+
+    /**
+     * Reads the full content as a [Bitmap].
+     */
+    suspend fun readAsBitmap(): ResourceTry<Bitmap> =
+        read().mapCatching {
```
mickael-menu, line 94: I would implement these functions as extensions on `Resource`, to make it clear that they are not part of the interface contract but helpers.

readium/shared/src/main/java/org/readium/r2/shared/resource/ZipContainer.kt
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * A [Container] representing a Zip archive.
+ */
+interface ZipContainer : Container {
+
+    interface Entry : Container.Entry {
+
+        /**
+         *  Compressed data length.
```
mickael-menu, line 31:
```suggestion
         * Compressed data length.
```

readium/shared/src/main/java/org/readium/r2/shared/resource/ZipContainer.kt
```diff
@@ -0,0 +1,188 @@
...
+    override suspend fun close() {
+        tryOrLog {
+            withContext(Dispatchers.IO) {
+                archive.close()
+            }
+        }
+    }
+}
+
+class DefaultArchiveFactory : ArchiveFactory {
```
mickael-menu, line 158: I would move this to `Factories.kt`, or better a dedicated `ArchiveFactory.kt` file.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/PublicationParser.kt
```diff
@@ -6,15 +6,26 @@
...
 /**
  *  Parses a Publication from an asset.
  */
 interface PublicationParser {
 
+    data class Asset(
+        val name: String,
+        val mediaType: MediaType,
+        val fetcher: Fetcher
+    )
```
mickael-menu, line 27: This looks very similar to `ContentProtection.Asset`. Is it something that we should refactor into a `PublicationAsset` (with the `onCreatePublication` closure)?

readium/streamer/src/main/java/org/readium/r2/streamer/parser/audio/AudioParser.kt
```diff
@@ -25,18 +28,31 @@ import org.readium.r2.streamer.parser.PublicationParser
...
-            return null
+        if (!asset.mediaType.matches(MediaType.ZAB) && !asset.mediaType.isAudio)
+            return Try.failure(PublicationParser.Error.FormatNotSupported())
 
-        val readingOrder = asset.fetcher.links()
-            .filter { link -> with(File(link.href)) { lowercasedExtension in audioExtensions && !isHiddenOrThumbs } }
-            .sortedBy(Link::href)
-            .toMutableList()
+        val readingOrder =
+            if (asset.mediaType.matches(MediaType.ZAB)) {
```
mickael-menu, line 40: Aren't we always in a ZAB case? Unless it's a single audio file, but in this case the sorting doesn't matter.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/image/ImageParser.kt
```diff
@@ -26,18 +24,31 @@ import org.readium.r2.streamer.parser.PublicationParser
...
-            return null
+        if (!asset.mediaType.matches(MediaType.CBZ) && !asset.mediaType.isBitmap)
+            return Try.failure(PublicationParser.Error.FormatNotSupported())
 
-        val readingOrder = asset.fetcher.links()
-            .filter { !File(it.href).isHiddenOrThumbs && it.mediaType.isBitmap }
-            .sortedBy(Link::href)
-            .toMutableList()
+        val readingOrder =
+            if (asset.mediaType.matches(MediaType.CBZ)) {
```
mickael-menu, line 36: Same comment as for the ZABs.

readium/streamer/src/main/java/org/readium/r2/streamer/PublicationFactory.kt
```diff
@@ -0,0 +1,267 @@
...
+     *   presenting dialogs.
+     * @param onCreatePublication Transformation which will be applied on the Publication Builder.
+     *   It can be used to modify the [Manifest], the root [Fetcher] or the list of service
+     *   factories of the [Publication].
+     * @param warnings Logger used to broadcast non-fatal parsing warnings.
+     * @return A [Publication] or a [Publication.OpeningException] in case of failure.
+     */
+    suspend fun open(
+        asset: Asset,
+        drmScheme: String? = null,
```
mickael-menu, line 119:
```suggestion
        contentProtectionScheme: ContentProtection.Scheme? = null,
```

readium/streamer/src/main/java/org/readium/r2/streamer/PublicationFactory.kt
```diff
@@ -0,0 +1,267 @@
...
+                it.open(asset, drmScheme, credentials, allowUserInteraction, sender)
+            }?.getOrElse { return Try.failure(it) }
+            ?: return Try.failure(Publication.OpeningException.Forbidden())
+
+        val parserAsset = PublicationParser.Asset(
+            protectedAsset.name,
+            protectedAsset.mediaType,
+            protectedAsset.fetcher
+        )
+
```
mickael-menu, line 180: Seeing this comforts me in the idea of having a shared `PublicationAsset` type. That would simplify the chaining of `unlock(Asset) -> PublicationAsset` and then `open(PublicationAsset) -> Publication`.

readium/streamer/src/main/java/org/readium/r2/streamer/PublicationFactory.kt
```diff
@@ -0,0 +1,267 @@
...
+    private fun Publication.addLegacyProperties(mediaType: MediaType?) {
+        @Suppress("DEPRECATION")
+        type = mediaType.toPublicationType()
+
+        if (mediaType == MediaType.EPUB)
+            setLayoutStyle()
+    }
+}
+
+internal fun MediaType?.toPublicationType(): Publication.TYPE =
```
mickael-menu, line 260: I guess we don't need this anymore, as `Publication.TYPE` is deprecated with error level?

readium/streamer/src/main/java/org/readium/r2/streamer/ParserAssetFactory.kt
```diff
@@ -0,0 +1,126 @@
...
+    ): Try<PublicationParser.Asset, Publication.OpeningException> {
+        return when (asset) {
+            is Asset.Container ->
+                Try.success(createFetcherForContainer(asset.container, asset.mediaType, asset.name))
+            is Asset.Resource ->
+                createFetcherForResource(asset.resource, asset.mediaType, asset.name)
+        }
+    }
+
+    private fun createFetcherForContainer(
```
mickael-menu, line 44: The name of this method and the following looks incorrect?

readium/streamer/src/main/java/org/readium/r2/streamer/ParserAssetFactory.kt
```diff
@@ -0,0 +1,126 @@
...
+            createFetcherForManifest(resource, assetName)
+        } else {
+            createFetcherForContent(resource, mediaType, assetName)
+        }
+
+    private suspend fun createFetcherForManifest(
+        resource: Resource,
+        assetName: String
+    ): Try<PublicationParser.Asset, Publication.OpeningException> {
+        val manifest = resource.readAsRwpm(packaged = false)
```
mickael-menu, line 68: What was the rationale for converting a manifest to a package asset here instead of in the `ReadiumWebPubParser` directly?

I'm concerned that this is not very extensible (e.g. if there's a format similar to RWPM that needs to be supported, W3C Web Publications for example). It also feels a bit weird to have RWPM specific code here.

readium/lcp/src/main/java/org/readium/r2/lcp/LcpContentProtection.kt
```diff
@@ -7,94 +7,192 @@
...
+        sender: Any?
+    ): Try<ContentProtection.Asset, Publication.OpeningException>? {
+        if (drmScheme != scheme.uri) {
+            return null
+        }
+
+        return when (asset) {
+            is Asset.Container -> openPublication(asset, credentials, allowUserInteraction, sender)
+            is Asset.Resource -> openLicense(asset, credentials, allowUserInteraction, sender)
+        }
```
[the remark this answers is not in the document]
qnga: I know that it is not completely obvious from the architecture, but this is not supposed to happen. The `AssetRetriever` is the only way you should get an asset and it is responsible for transforming resources into containers in case of archives.
Actually, it's slightly visible through the `Asset.assetType` property. It's forbidden to be a `Resource` if you have the `Archive` `AssetType`.

readium/shared/src/main/java/org/readium/r2/shared/asset/AssetType.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Use of this source code is governed by the BSD-style license
+ * available in the top-level LICENSE file of the project.
+ */
+
+package org.readium.r2.shared.asset
+
+enum class AssetType {
+
+    /**
+     * A simple resource.
```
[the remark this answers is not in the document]
qnga: I'm not sure. I've used `simple` as the opposite of `complex`: the resource is not only single, but also has a simple structure: it doesn't contain several resources.

readium/shared/src/main/java/org/readium/r2/shared/asset/AssetRetriever.kt
```diff
@@ -0,0 +1,351 @@
...
+import org.readium.r2.shared.util.mediatype.*
+import org.readium.r2.shared.util.toUrl
+
+class AssetRetriever(
+    private val resourceFactory: ResourceFactory,
+    private val containerFactory: ContainerFactory,
+    private val archiveFactory: ArchiveFactory,
+    contentResolver: ContentResolver,
+    sniffers: List<Sniffer>
+) {
```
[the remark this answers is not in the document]
qnga: I think I wanted to force apps to be aware of what they do. We could have an alternate constructor with only a `Context` parameter though. That makes sense for apps that want to support only local files.

readium/shared/src/main/java/org/readium/r2/shared/resource/ContentResource.kt
```diff
@@ -0,0 +1,148 @@
...
+    private suspend fun readRange(range: LongRange): ResourceTry<ByteArray> =
+        withStream {
+            withContext(Dispatchers.IO) {
+                val skipped = it.skip(range.first)
+                check(skipped == range.first)
+                val length = range.last - range.first + 1
+                it.read(length)
+            }
+        }
+
```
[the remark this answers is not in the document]
qnga: Possibly. It's not so clear though, because this is only reading local files after all.

readium/shared/src/main/java/org/readium/r2/shared/resource/FileChannelResource.kt
```diff
@@ -0,0 +1,102 @@
...
+import java.nio.channels.Channels
+import java.nio.channels.FileChannel
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import org.readium.r2.shared.error.Try
+import org.readium.r2.shared.extensions.*
+import org.readium.r2.shared.extensions.read
+import org.readium.r2.shared.extensions.readFully
+
+class FileChannelResource(
```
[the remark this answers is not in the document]
qnga: Yes, I've kept it for further investigation with `content` schemes. I made it internal for now.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/audio/AudioParser.kt
qnga: The filtering matters a bit more, though not so much. Anyway, I feel like what we're doing is a bit clearer this way.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/image/ImageParser.kt
qnga: Here filtering might be critical if the `Fetcher` is not aware of the single file media type.

readium/streamer/src/main/java/org/readium/r2/streamer/PublicationFactory.kt
qnga: In the next PR only!

readium/streamer/src/main/java/org/readium/r2/streamer/ParserAssetFactory.kt
qnga: Legacy, thanks.

readium/streamer/src/main/java/org/readium/r2/streamer/ParserAssetFactory.kt
qnga: In `ReadiumWebPubParser` it was too late to create a `Fetcher` because `ContentProtection` might have modified it before. It's no longer the case: using manifests seems incompatible with `ContentProtection`s unless a `ContentProtection` itself implements this stuff.

I feel like this might deserve more thought though, let's postpone this change.

readium/shared/src/main/java/org/readium/r2/shared/publication/protection/ContentProtection.kt
qnga: Sounds good!

readium/shared/src/main/java/org/readium/r2/shared/asset/AssetRetriever.kt
```diff
@@ -0,0 +1,351 @@
...
+        mediaType: String? = null,
+        fileExtension: String? = null
+    ): Asset? {
+        return retrieve(url, listOfNotNull(mediaType), listOfNotNull(fileExtension))
+    }
+
+    /**
+     * Retrieves an asset from a Url.
+     */
+    private suspend fun retrieve(
```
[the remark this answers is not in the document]
qnga: I don't think so!

readium/shared/src/main/java/org/readium/r2/shared/resource/BytesResource.kt
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package org.readium.r2.shared.resource
+
+import kotlinx.coroutines.runBlocking
+import org.readium.r2.shared.error.Try
+import org.readium.r2.shared.extensions.coerceIn
+import org.readium.r2.shared.extensions.requireLengthFitInt
+
+sealed class BaseBytesResource(
```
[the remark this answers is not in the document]
qnga: It's been sealed since long to prevent subclassing. I thought it as an internal tool, not a public API. Apps can manage on their own if there's a need. The implementation is so trivial that I'm sure it still makes sense even internally nowadays.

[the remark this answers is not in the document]
qnga: Mmm, I confused this class with the `Fetcher.Resource` one. Actually,I just applied the same pattern.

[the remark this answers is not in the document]
mickael-menu: I'm fine keeping this as an internal tool, I'm just a bit confused about the sealed class. If it's just to prevent subclassing, unless we add `open` it's already the case?

readium/streamer/src/main/java/org/readium/r2/streamer/parser/audio/AudioParser.kt
mickael-menu: Fine by me.

readium/streamer/src/main/java/org/readium/r2/streamer/ParserAssetFactory.kt
mickael-menu: OK, I'll add an entry in the V3 GH project

readium/shared/src/main/java/org/readium/r2/shared/resource/BytesResource.kt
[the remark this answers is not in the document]
qnga: We need to subclass internally. Sealing a class is the way to enable only internal subclassing.

mickael-menu (approved):
Thank you for making the changes, it's coming along nicely 👍

PR #364, opened 2023-07-13
Remove Fuel and Kovenant dependencies

stevenzeck:
This removes the dependencies on the Fuel and Kovenant libraries. All code that used these libraries has been removed, with all but one of the user-facing functions having previously been deprecated. The one that was not deprecated is [Response.sniffMediaType](https://github.com/readium/kotlin-toolkit/blob/57bc32b7528b932b7165bd913f29741916950231/readium/shared/src/main/java/org/readium/r2/shared/util/mediatype/Extensions.kt#L45-L68).

If preferred, Fuel can go back in while we deprecate the function to give users time to update their own code. It can't be rewritten since it's an extension function of Fuel's own `Response`. Doing a search of "fuel sniffmediatype" here on GitHub only returned results of the toolkit itself (and people's own fork of it). Of course that doesn't cover anything outside of GitHub.


qnga:
I would restore fuel, `Response.sniffMediaType`, as well as all of the other symbols that were only deprecated with a warning level so far. That's the general process for v3. As many other things, we'll be able to remove them as soon as v3 is released for the next version.

mickael-menu:
I would expect nobody uses this API, although the most user-friendly path would be to keep Fuel as deprecated `level = ERROR` and remove it completely in 3.1 as @qnga suggested.

Or maybe removing it completely but adding a mention in the migration guide would be enough?

mickael-menu:
Thank you @stevenzeck for tackling this!

I would expect nobody uses this API, although the most user-friendly path would be to keep Fuel as deprecated `level = ERROR` and remove it completely in 3.1 as @qnga suggested.

Or maybe removing it completely but adding a mention in the migration guide would be enough?

qnga:
That makes sense. There will be no surprise for users if we make clear in the migration guide that fuel support is removed.

qnga (approved):
Looks good to me.

mickael-menu (approved):
Thank you @stevenzeck, this is much appreciated 🙏

PR #365, opened 2023-07-14
Prepare v3

qnga:
- Prefix all resources with readium_
- Remove Ref<Publication> from the Publication.Context
- Remove ReadingProgression.TTB/BTT/AUTO.
- Remove the Parcelable conformance on Manifest, and deprecate the PublicationRepository.
- Remove all the Activity navigators.
- Remove older settings API.
- Remove the HTTP server.
- Remove highlight.js


readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorViewModel.kt
```diff
@@ -178,8 +161,7 @@ internal class EpubNavigatorViewModel(
     // Serving resources
 
     val baseUrl: String =
-        baseUrl?.let { it.removeSuffix("/") + "/" }
-            ?: publication.linkWithRel("self")?.href
+        publication.linkWithRel("self")?.href
             ?: WebViewServer.publicationBaseHref
```
qnga, line 166: I'm not sure about this. Why should we deal with two different cases?

mickael-menu: We currently support two cases:

* A remote web publication, which will have a `self` link in its manifest pointing to itself on the HTTP server. In this case, we won't use the `WebViewServer`.
* A local publication, e.g. parsed from an EPUB, which won't have any `self` link. In this case the publication is served through the `WebViewServer`.

mickael-menu (approved):
LGTM, I'm going to merge it and open a new PR for additional fixes.

PR #368, opened 2023-07-24
Updating dependencies, SDKs, and more

stevenzeck:
**What's been updated**

1. Update all dependencies to latest versions.
2. Update compile and target SDKs to 34.
3. Update Gradle to 8.0, and AGP to 8.1.0.
4. Use Java 17 (required by AGP update).

**Important notes**

1. [KSP](https://developer.android.com/build/migrate-to-ksp) has replaced Kapt for Room. There is an [issue](https://issuetracker.google.com/issues/236612358) with Room 2.5.2 and Kotlin 1.9.0 that won't be fixed until 2.5.3. One of the workarounds in the issue is to use KSP instead of Kapt. Kapt is in maintenance mode and KSP offers significant performance improvements.
2. With Kotlin 1.8, we don't need to declare `kotlin-stdlib-jdk7` or 8 as they have been merged into a straightforward `kotlin-stdlib`. https://kotlinlang.org/docs/whatsnew18.html#updated-jvm-compilation-target.
3. android.nonTransitiveRClass defaults to true with the new AGP version. Because of this, when accessing IDs from external libraries, you have to prefix it with the namespace. For example, `R.id.search_close_btn` is now `androidx.appcompat.R.id.search_close_btn`.
4. If `BuildConfig` is needed for individual modules, `buildConfig = true` must be added to the `buildFeatures` block in build.gradle.kts for that module.

**What else changed**

1. When targeting Android SDK 34 you must add `android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK` in order for TTS or Audiobook playback to work.


stevenzeck:
Thinking the first two issues are from updating Java. Android Gradle Plugin 8.0+ requires Java 17 or higher, so have to downgrade that too if necessary.

stevenzeck:
@mickael-menu @qnga Please review, especially the Kotlin file changes.

I tested everything except for LCP. But please test with your devices too.

test-app/src/main/AndroidManifest.xml
```diff
@@ -11,6 +11,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK" />
```
mickael-menu, line 14: Is this required when targeting API 34? Then we need to update the user guide too: https://github.com/readium/kotlin-toolkit/blob/f911a2cae938868d934a4d24ddb1794c3787edf8/docs/guides/media-navigator.md?plain=1#L125

readium/navigator/build.gradle.kts
```diff
@@ -70,7 +71,7 @@ dependencies {
     implementation(libs.androidx.viewpager2)
     implementation(libs.androidx.webkit)
     // Needed to avoid a crash with API 31, see https://stackoverflow.com/a/69152986/1474476
-    implementation("androidx.work:work-runtime-ktx:2.7.1")
+    implementation("androidx.work:work-runtime-ktx:2.8.1")
```
mickael-menu, line 74: After a quick test on 31 emulator, I think this is not needed anymore. At least with the media3 navigators, we will deprecate the media1 and media2 ones.

settings.gradle.kts
```diff
@@ -18,13 +18,14 @@ pluginManagement {
...
+        id("com.android.library") version ("8.0.2")
         id("io.github.gradle-nexus.publish-plugin") version ("1.1.0")
-        id("org.jetbrains.dokka") version ("1.7.20")
-        id("org.jetbrains.kotlin.android") version ("1.7.20")
-        id("org.jetbrains.kotlin.plugin.serialization") version ("1.7.10")
+        id("org.jetbrains.dokka") version ("1.8.20")
+        id("org.jetbrains.kotlin.android") version ("1.9.0")
+        id("org.jetbrains.kotlin.plugin.serialization") version ("1.9.0")
         id("org.jlleitschuh.gradle.ktlint") version ("11.0.0")
+        id("com.google.devtools.ksp") version ("1.9.0-1.0.12")
```
mickael-menu, line 28:
```suggestion
        // Make sure to align with the Kotlin version.
        // See https://github.com/google/ksp/releases
        id("com.google.devtools.ksp") version ("1.9.0-1.0.12")
```

mickael-menu (commented):
Thanks @stevenzeck, I tested again the various formats with LCP too.

readium/adapters/pspdfkit/pspdfkit-navigator/src/main/java/org/readium/adapters/pspdfkit/navigator/PsPdfKitDocumentFragment.kt
```diff
@@ -192,9 +192,9 @@ internal class PsPdfKitDocumentFragment(
         }
 
         private val allowedTextSelectionItems = listOf(
-            R.id.pspdf__text_selection_toolbar_item_share,
-            R.id.pspdf__text_selection_toolbar_item_copy,
-            R.id.pspdf__text_selection_toolbar_item_speak
+            com.pspdfkit.R.id.pspdf__text_selection_toolbar_item_share,
+            com.pspdfkit.R.id.pspdf__text_selection_toolbar_item_copy,
+            com.pspdfkit.R.id.pspdf__text_selection_toolbar_item_speak
```
qnga, line 197: Does this change have consequences? Maybe these resources cannot be overridden any more.

test-app/src/main/AndroidManifest.xml
stevenzeck: Yes it's required for Android 34. Updated the media navigator doc and the migration guide.

readium/adapters/pspdfkit/pspdfkit-navigator/src/main/java/org/readium/adapters/pspdfkit/navigator/PsPdfKitDocumentFragment.kt
stevenzeck: [This](https://github.com/readium/kotlin-toolkit/pull/368#discussion_r1273524776) is actually because android.nonTransitiveRClass defaults to true with the [updated AGP version](https://developer.android.com/build/releases/gradle-plugin#default-changes).

Another fix would be to set that to false in gradle.properties, then we can continue using `R.id.something` instead of `package.namespace.R.id.something`. There is only one other place this was necessary and it was in the testapp.

stevenzeck:
Looks to be an issue with ExoPlayer in which the test app requires `android.permission.POST_NOTIFICATIONS`. https://github.com/google/ExoPlayer/issues/10884

readium/adapters/pspdfkit/pspdfkit-navigator/src/main/java/org/readium/adapters/pspdfkit/navigator/PsPdfKitDocumentFragment.kt
mickael-menu: I confirm that we can still override PSPDFKit strings with these changes, from an app's module. I tried with this:

```xml
<string name="pspdf__action_menu_copy">Test</string>
```

test-app/src/main/AndroidManifest.xml
mickael-menu: Thank you!

mickael-menu:
@stevenzeck The ExoPlayer DownloadService is not used in the toolkit, so it should not be required. Did you encounter an issue with the Test App?

mickael-menu:
So apparently we need this permission even without using the `DownloadService`, because otherwise the linting step fails. See this comment: https://github.com/google/ExoPlayer/issues/10884#issuecomment-1450094932

mickael-menu (approved):

PR #371, opened 2023-07-27
Cleanup work from #368

stevenzeck:
**What was done**

1. androidx-compose-theme-adapter has been deprecated in favor of the one in [Accompanist](https://github.com/google/accompanist). Removed the former and added the latter.
2. Updated a few lingering libraries.
3. Replaced remaining uses of `launchWhenStarted` and `launchWhenResumed` with `launch` and `repeatOnLifecycle`.
4. Used a `Builder` for DeviceInfo in TtsSessionAdapter, as the constructor was deprecated.
5. Replaced onCreateOptionsMenu/onOptionsItemSelected with MenuHost/MenuProvider.
6. Changed CatalogFragment/CatalogViewModel to use EventChannel instead of MutableLiveData.
7. Added a background color back into search result headings.


stevenzeck:
@mickael-menu If you have any ideas for the second point under Issues, please let know. Also if you know of an OPDS feed with facets to make sure the toolbar menu still works, that would also help.

mickael-menu (approved):
Thank you for handling the grunt work @stevenzeck!

I adjusted the colors on the search fragment to work in dark mode and avoid hard-coded colors.
Screenshot_1690536656
Screenshot_1690536664

PR #373, opened 2023-08-01
Replace the `Fetcher` by `Container` and refactor media type sniffing

mickael-menu:
#### Shared

 * `Publication.localizedTitle` is nullable, as we cannot guarantee that all publication sources offer a title.
* The `MediaType` sniffing helpers are deprecated in favor of `MediaTypeRetriever` (for media type and file extension hints and raw content) and `AssetRetriever` (for URLs). 

#### Streamer

 * The `Fetcher` interface was deprecated in favor of the `Container` one in `readium-shared`.

## Notes

* `Href` API was revamped and now supports `file://` schemes.
* Removed `Asset.name`, `Container.name` and `Resource.name`.


readium/lcp/src/main/java/org/readium/r2/lcp/LcpDecryptor.kt
```diff
@@ -17,30 +17,40 @@ import org.readium.r2.shared.extensions.coerceFirstNonNegative
...
-        val link = resource.link()
-        val encryption = link.properties.encryption
-        if (encryption == null || encryption.scheme != "http://readium.org/2014/01/lcp")
-            return@LazyResource resource
-
-        when {
-            license == null -> FailureResource(link, Resource.Exception.Forbidden())
-            link.isDeflated || !link.isCbcEncrypted -> FullLcpResource(resource, license)
-            else -> CbcLcpResource(resource, license)
+internal class LcpDecryptor(
```
mickael-menu, line 33: `LcpDecryptor` don't require the encryption metadata from the input `Link` anymore. Instead, we have an explicit `retrieveEncryption` dependency which will return the encryption metadata associated with a source URL.

readium/shared/src/main/java/org/readium/r2/shared/resource/Resource.kt
```diff
@@ -29,24 +30,37 @@ public typealias ResourceTry<SuccessT> = Try<SuccessT, Resource.Exception>
...
+        public companion object {
+            public inline operator fun invoke(build: Builder.() -> Unit): Properties =
+                Properties(Builder().apply(build))
+        }
+
+        public inline fun copy(build: Builder.() -> Unit): Properties =
+            Properties(Builder(this).apply(build))
+
+        public class Builder(properties: Map<String, Any> = emptyMap())
+            : MutableMap<String, Any> by properties.toMutableMap()
```
mickael-menu, line 62: This allows to define type-safe extensions:

```kotlin
 private const val archiveKey = "archive"

 public val Resource.Properties.archive: ArchiveProperties?
     get() = (this[archiveKey] as? Map<*, *>)
         ?.let { ArchiveProperties.fromJSON(JSONObject(it)) }

 public var Resource.Properties.Builder.archive: ArchiveProperties?
     get() = (this[archiveKey] as? Map<*, *>)
         ?.let { ArchiveProperties.fromJSON(JSONObject(it)) }
     set(value) {
         if (value == null) {
             remove(archiveKey)
         } else {
             put(archiveKey, value.toJSON())
         }
     }
```

which can be used like that:

```kotlin
val props = Properties {
    archive = ArchiveProperties(...)
}


val props2 = props.copy {
    archive = ArchiveProperties()
}

props2.archive
```

mickael-menu:
The CBZ parser tests won't pass until the media type issue is fixed.

qnga (approved):

Issue #124, opened 2022-05-20
Impossible to open a file when linear="no"


civodulab:
Impossible to open a file whose `itemref` attribute is `linear="no"`.
However it works in Thorium.

mickael-menu:
Yes non linear resources are not yet supported, mostly because the correct behavior is unclear. See this discussion https://github.com/readium/architecture/discussions/143

What do you personally expect should happen?
* Opening a pop-up with the non linear resource inside.
* Replacing the currently visible resource by the non linear one, then how do you go back?

Do you have a sample book with `linear="no"`?

civodulab:
I am working on a book for visually impaired youngsters. There is a rather long notice explaining how the options work. I wanted to hide it to avoid too many pages to turn. This notice will only be read once or twice. On my configuration page I have a link to this manual and on the page of this manual I have a link to the configuration page.

Since I saw that the option existed I wanted to use it.

> What do you personally think should happen?
> Open a pop-up with the non-linear resource inside.
> Replace the currently visible resource with the non-linear resource, then how do you go back?

Replace the currently visible resource with the non-linear resource.

I made a very quick epub with a non-linear resource.
[test_linear_no.zip](https://github.com/readium/kotlin-toolkit/files/8752540/test_linear_no.zip)
.

mickael-menu:
Thank you, such samples are very useful.

This feature is [a real can of worms](https://github.com/readium/architecture/discussions/143) so I doubt we will get a good implementation soon. But maybe we can work out a limited solution in the meantimes, as it's completely broken right now.

qnga:
Closing as we've implemented a way for apps to handle non-linear resources and it's not clear if we'll do more.

PR #375, opened 2023-08-18
Extend Navigator interface with shouldJumpToLink method

chrfalch:
When the user clicks an internal link in the Navigator the link might be to a resource not present in the reading order - example would be a note that was added with linear=no in the epub (Impossible to open a file when linear="no" https://github.com/readium/kotlin-toolkit/issues/124).

This PR implements a way to intercept such a link click - and also for the listener to tell the Navigator not to navigate to the link - meaning that the app can implement handling the link itself (for example showing a popup with the contents of the link like iBooks do).

The implementation is done by extending the Navigator interface with a method called `shouldJumpToLink` that returns true by default and is only called when a link is clicked (when someone calls the `go` method with a link object - which is done when a link internal to the epub is clicked).


readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorViewModel.kt
```diff
@@ -205,7 +206,9 @@ internal class EpubNavigatorViewModel(
         val href = url.toString()
         val link = internalLinkFromUrl(href)
         if (link != null) {
-            _events.send(Event.GoTo(link))
+            if (listener?.onShouldJumpToLink(link) == true) {
+                _events.send(Event.GoTo(link))
```
qnga, line 210: For the sake of consistency, I guess we could have a specialized event and do the check in `EpubNavigatorFragment`.

chrfalch: So that the ViewModel doesn't need to "know" about the listener? The Event.GoTo is only used here, so I don't know if we'd need to have a different one? 

I think the idea here (from @mickael-menu) was to ensure that if anyone else sent this event we would avoid calling the `onShouldJumpToLink` callback by making this "private" to the ViewModel. 

My first thought was to call `onShouldJumpToLink` in the EPubNavigatorFragment when it handles events.

qnga: Ok! I would follow your first thought, renaming `Event.GoTo` to `Event.InternalLinkClicked`, both approaches are fine though.

mickael-menu: `Event` is poorly named (my fault...), it's more like commands that the view model send to the navigator. Initially `GoTo` was meant to be a general go to command that could be used for other reasons the view model might want to move. But it looks like it's only being used for internal links, so renaming it to `Event.OpenInternalLink` would be fine. (And maybe renaming `Event` to `Command`?).

chrfalch: If we call it `OpenInternalLink` - should we still delegate to the ViewModel to call the listener's `onShouldOpenLink`()?

I'll leave `Event` -> `Command` out for now :)

qnga: All of that is private, so it doesn't matter much after all.

qnga (approved):
I'm fine with that.

mickael-menu (approved):

PR #376, opened 2023-08-18
Add reading order as parameter to EPubFragment factory / constructor

chrfalch:
To be able to show resources not part of the reading order (like notes, see Impossible to open a file when linear="no" #124) it is helpful to be able to provide a special readingOrder list when opening a publication in the navigator.

This commit adds readingOrder as an optional parameter with the default value of `Publication.readingOrder` to the epub factory and the fragment.

In addition the EPubFragment is now using the provided readingOrder instead of publication.readingOrder in its implementation.


readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
```diff
@@ -306,13 +307,15 @@ class EpubNavigatorFragment internal constructor(
         _binding = ActivityR2ViewpagerBinding.inflate(inflater, container, false)
         val view = binding.root
 
+        // FIXME: If readingOrder is provided as input to the fragment, we might not want
+        // to map through the publication's readingOrder:
         positionsByReadingOrder = runBlocking { publication.positionsByReadingOrder() }
         positions = positionsByReadingOrder.flatten()
         publicationIdentifier = publication.metadata.identifier ?: publication.metadata.title
```
mickael-menu, line 314: We talked about this issue with @HadrienGardeur and @qnga and came up with this consensus:

Positions and total progression only make sense in the context of the publication's actual reading order. Therefore when provided with a different reading order, we should assume the positions list is empty, and also not compute the `totalProgression` when calculating the current locator.

(We might revisit this decision when introducing the guided navigations APIs, but better add more metadata than having to remove them at some point)

To avoid comparing the provided `readingOrder` to the publication one, I think you can just set the constructor parameter as a nullable `List<Link>?` and ignore the positions and total progression when it is not null.

readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
```diff
@@ -1035,21 +1037,27 @@ class EpubNavigatorFragment internal constructor(
...
             config: Configuration = Configuration(),
             initialPreferences: EpubPreferences = EpubPreferences()
         ): FragmentFactory =
             createFragmentFactory {
                 EpubNavigatorFragment(
-                    publication, baseUrl, initialLocator, initialPreferences,
+                    publication, baseUrl, initialLocator,
+                    readingOrder = readingOrder ?: publication.readingOrder,
+                    positionsByReadingOrder = if (readingOrder == null)
+                        Collections.emptyList() else runBlocking { publication.positionsByReadingOrder() },
```
qnga, line 1059: Do we really want to expose a `positionsByReadingOrder` argument? I feel like our approach is fine for now.

readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
```diff
@@ -306,13 +308,13 @@ class EpubNavigatorFragment internal constructor(
         _binding = ActivityR2ViewpagerBinding.inflate(inflater, container, false)
         val view = binding.root
 
-        positionsByReadingOrder = runBlocking { publication.positionsByReadingOrder() }
+        // If we have a custom reading order we have to skip calculating positions.
```
qnga, line 311: Isn't this comment outdated?

chrfalch: Yeah - removed now!

readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
chrfalch: Agree - this is more me not being 100% sure on how to solve the constructor arguments in Kotlin... Maybe one of you can suggest something here?

mickael-menu: Yes you could take only the `readingOrder: List<Link>?` in the internal constructor and initialize the internal properties. This way you won't have to do the same thing in `EpubNavigatorFactory` either.

```kotlin
private val readingOrder: List<Link> = readingOrder ?: publication.readingOrder

private val positionsByReadingOrder: List<List<Locator>> =
    if (readingOrder == null) emptyList()
    else runBlocking { publication.positionsByReadingOrder() }
```

chrfalch: For `positionsByReadingOrder` to be able to test for null on readingOrder, I kept `readingOrder` as List<Link>? = null in the constructor.

Instead we check for null when we try to use `readingOrder` to not break `positionsByReadingOrder`.

chrfalch: Wow, good catch, and sorry 🙀 - I was a bit quick here!! 

I'll see if I can find out about the constructor.

chrfalch: But if I declare the property with the same name - how would positionsByReadingOrder be able to check if the constructor argument for readingOrder was null? @qnga

chrfalch: Seems like that is automatically solved in the positionsByReadingOrder property - so then we should be all good. Pushed a few updates.

readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFactory.kt
```diff
@@ -47,6 +52,9 @@ class EpubNavigatorFactory(
             publication = publication,
             baseUrl = null,
             initialLocator = initialLocator,
+            readingOrder = readingOrder ?: publication.readingOrder,
+            positionsByReadingOrder = if (readingOrder == null)
+                Collections.emptyList() else runBlocking { publication.positionsByReadingOrder() },
```
qnga, line 57: This is no longer needed.

chrfalch: Thanks, removed now. (Starting to get a bit embarrassed now over my Kotlin short-comings)..

readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFactory.kt
```diff
@@ -47,6 +52,7 @@ class EpubNavigatorFactory(
             publication = publication,
             baseUrl = null,
             initialLocator = initialLocator,
+            readingOrder = readingOrder ?: publication.readingOrder,
```
qnga, line 55:
```suggestion
            readingOrder = readingOrder
```
If `readingOrder` is never null, positions will always be used in the fragment.

qnga (approved):
Looks good to me, though I didn't test it.

mickael-menu (approved):
Thank you @chrfalch and @qnga

PR #380, opened 2023-08-23
Simplify RWPM parsing for packaged publications

mickael-menu:
* Don't remove the `self` link anymore, when parsing a packaged publication. They are not supposed to contain a `self` link with an absolute URL.
* Fix file extensions in the LCP service.
* Restore visibility of `MediaSessionNavigator` properties (hiding them was a mistake when migrating to the explicit API mode).


qnga:
You can now revert the change in `ParserAssetFactory` and use again `asset.resource` instead of the parsed manifest.

qnga (approved):

PR #381, opened 2023-08-25
Add a download manager

qnga:
The `DownloadManager` component makes it possible to download files in the backgroud, even  when the app is not started.


readium/lcp/src/main/java/org/readium/r2/lcp/LcpPublicationRetriever.kt
```diff
@@ -0,0 +1,248 @@
...
+import org.readium.r2.lcp.license.model.LicenseDocument
+import org.readium.r2.shared.extensions.tryOrLog
+import org.readium.r2.shared.util.Try
+import org.readium.r2.shared.util.Url
+import org.readium.r2.shared.util.getOrElse
+import org.readium.r2.shared.util.mediatype.FormatRegistry
+import org.readium.r2.shared.util.mediatype.MediaType
+import org.readium.r2.shared.util.mediatype.MediaTypeRetriever
+
+private val Context.dataStore: DataStore<Preferences> by preferencesDataStore(
```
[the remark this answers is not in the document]
qnga: Not so easy to achieve. I switched to a file storage in `Context.noBackupFilesDir` instead.

readium/downloads/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
```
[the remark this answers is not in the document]
qnga: Same as regarding LCP licenses.

mickael-menu:
@CodiumAI-Agent /review

mickael-menu:
@CodiumAI-Agent /ask We can create multiple `LcpLicenseRetriever` with several listeners, but it's an issue as the downloads are handled globally and we might loose some events. Any suggestion?

readium/shared/src/main/java/org/readium/r2/shared/util/downloads/foreground/ForegroundDownloadManager.kt
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public override fun register(
+        requestId: DownloadManager.RequestId,
+        listener: DownloadManager.Listener
+    ) {
+        listeners.getOrPut(requestId) { mutableListOf() }.add(listener)
+    }
+
+    public override fun close() {
```
mickael-menu, line 134: Maybe we can cancel all the known requests in `close`, to let the opportunity to recover gracefully in the listeners?

readium/lcp/src/main/java/org/readium/r2/lcp/LcpPublicationRetriever.kt
```diff
@@ -0,0 +1,214 @@
...
+                            LcpException.wrap(
+                                Exception("Couldn't retrieve license from local storage.")
+                            )
+                        )
+                    }
+                    return
+                }
+            downloadsRepository.removeDownload(requestId.value)
+
+            val link = license.link(LicenseDocument.Rel.Publication)!!
```
mickael-menu, line 79: Since we know that there will always be a publication link in the contract, how about adding this in `LicenseDocument` to avoid the force unwrapping in the callers?

```kotlin
val LicenseDocument.publicationLink: Link get() = link(Rel.Publication)!!
```

readium/lcp/src/main/java/org/readium/r2/lcp/LcpPublicationRetriever.kt
```diff
@@ -0,0 +1,214 @@
...
+
+    private val listeners: MutableMap<RequestId, MutableList<Listener>> =
+        mutableMapOf()
+
+    public fun register(
+        requestId: RequestId,
+        listener: Listener
+    ) {
+        listeners.getOrPut(requestId) { mutableListOf() }.add(listener)
+        downloadManager.register(DownloadManager.RequestId(requestId.value), downloadListener)
```
mickael-menu, line 170: I think we will get duplicated calls if we have several listeners, as we register to the download manager again.
```suggestion
        if (listeners[requestId].size() == 1) {
            downloadManager.register(DownloadManager.RequestId(requestId.value), downloadListener)
        }
```

readium/lcp/src/main/java/org/readium/r2/lcp/LcpPublicationRetriever.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private inner class DownloadListener : DownloadManager.Listener {
+
+        override fun onDownloadCompleted(
+            requestId: DownloadManager.RequestId,
+            file: File
+        ) {
+            val lcpRequestId = RequestId(requestId.value)
+            val listenersForId = listeners[lcpRequestId].orEmpty()
```
mickael-menu, line 62: I believe this should be an error here if we don't have any LCP listeners recorded for this ID, as we could loose notification.

test-app/src/main/res/values/arrays.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string-array name="documentSelectorArray">
-        <item>Copy to app storage</item>
+        <item>Import to app storage</item>
```
mickael-menu, line 4: I think `Copy` was more consistent with the other options. All of them are **importing** publications, but the verb explains the way the publication file is accessed.

qnga: I introduced a subtle distinction in the `Bookshelf` that is reflected here: you can _add_ an asset to the database, or _import_ a publication from it to the app storage. In the latter case but not in the former, LCP publications will be downloaded and stored locally.

mickael-menu (approved):
LGTM, thank you 👌

PR #384, opened 2023-08-28
Fix/383 scroll to top

### LINKED ISSUE #383: Navigator places locator with anchor in the middle of the screen in scroll mode
When an EPub navigator is in scroll mode and navigating to a locator with an anchor (ex.: /OEBPS/chapter16.xhtml#endnote-022), the scroll position is calculated to place the target element in the middle of the screen. This can be an issue for footnotes in large indexes because it might not be clear which footnote that was clicked.

See same issue in swift-toolkit here:

https://github.com/readium/swift-toolkit/issues/328

chrfalch:
To make scrolling to locations in lists like footnotes etc the scrollToRect method now places the target element as the topmost element when scrolling.

Fixes Navigator places locator with anchor in the middle of the screen in scroll mode

Same implementation as for swift toolkit: https://github.com/readium/swift-toolkit/pull/329

Fixes #383


mickael-menu (approved):
Thank you @chrfalch!

PR #385, opened 2023-08-30
Fix encoding of `Href` query and fragment

mickael-menu:
Historically we percent-decoded HREFs because we can have an EPUB with resources containing, for example, a percent character. The way it was implemented caused an issue with `http://` HREF with a fragment or query param containing special characters (e.g. `%` or `#`).

To fix this, I adjusted the HREF parsing to not percent-decode the path but not the fragment and query params.


mickael-menu:
I'm preparing another PR that might significantly change `Href`.

mickael-menu:
Superseded by https://github.com/readium/kotlin-toolkit/pull/388

PR #390, opened 2023-09-14
Bugfixes for TTS

mickael-menu:
* Fix issue when computing a substring of a `Locator.Text`, were the char offsets were incorrect.
* Fix terminating the current TTS task.
    * Sometimes we received events for a previous task, which interrupted the current task too early. I fixed it by checking for the task ID first.

### DIFF
--- a/readium/navigator/src/main/java/org/readium/r2/navigator/media3/tts/TtsEngineFacade.kt
+++ b/readium/navigator/src/main/java/org/readium/r2/navigator/media3/tts/TtsEngineFacade.kt
@@ -26,8 +26,6 @@ internal class TtsEngineFacade<S : TtsEngine.Settings, P : TtsEngine.Preferences
         engine.setListener(listener)
     }
 
-    private var currentTask: UtteranceTask<E>? = null
-
     val voices: Set<V>
         get() = engine.voices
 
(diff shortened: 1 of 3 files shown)


readium/navigator/src/main/java/org/readium/r2/navigator/media3/tts/TtsEngineFacade.kt
```diff
@@ -45,54 +43,44 @@ internal class TtsEngineFacade<S : TtsEngine.Settings, P : TtsEngine.Preferences
...
         val continuation: CancellableContinuation<E?>,
         val onRange: (IntRange) -> Unit
     )
 
+    private fun getTask(id: TtsEngine.RequestId) =
+        currentTask?.takeIf { it.requestId == id }
+
+    private fun popTask(id: TtsEngine.RequestId) =
+        getTask(id)
+            ?.also { currentTask = null }
```
qnga, line 59: Is that the important thing? I can't see any difference in the listener, except this. Which call to `currentTask` happened to receive the wrong task?

mickael-menu: Yes. It was `onFlushed`, which was called with the ID of the previous task than the `currentTask` one.

As we didn't check the id before setting `currentTask = null`, it was breaking the playback.

qnga (approved):
Ok, that looks correct to me.

PR #398, opened 2023-09-26
Fix configuration change crash with `PdfNavigatorFragment`

mickael-menu:
* Fix crash during configuration changes with PDFium and PSPDFKit.
* The `PdfDocumentFactory` now returns a `ResourceTry<PdfDocument>` instead of throwing an exception in case of an error.
* The `PdfEngineProvider` now provides a `FragmentFactory` for the `PdfDocumentFragment`.
* Added a listener and customization point for PSPDFKit.
* During the fragment lifecycle, the `PdfNavigatorFragment` is the source of truth for the current page and settings. When restoring the fragment, the `PdfDocumentFragment` is recreated from the last known page index and settings stored in the view model. This is to simplify the implementations of `PdfDocumentFragment`, which don't need to save and restore any state.


readium/adapters/pdfium/pdfium-navigator/src/main/java/org/readium/adapters/pdfium/navigator/PdfiumEngineProvider.kt
```diff
@@ -22,19 +24,22 @@ import org.readium.r2.shared.publication.Publication
...
-            initialPageIndex = input.initialPageIndex,
-            settings = input.settings,
-            appListener = listener,
-            navigatorListener = input.listener
-        )
+) : PdfEngineProvider<PdfiumDocumentFragment, PdfiumDocumentFragment.Listener, PdfiumSettings, PdfiumPreferences, PdfiumPreferencesEditor> {
+
+    override fun createDocumentFragmentFactory(
+        input: PdfDocumentFragmentInput<PdfiumDocumentFragment.Listener, PdfiumSettings>
+    ): SingleFragmentFactory<PdfiumDocumentFragment> =
```
mickael-menu, line 32: We return a `SingleFragmentFactory` which is a custom implementation of `FragmentFactory` able to instantiate a new instance without knowing the class of the fragment.

readium/adapters/pspdfkit/pspdfkit-navigator/src/main/java/org/readium/adapters/pspdfkit/navigator/PsPdfKitDocumentFragment.kt
```diff
@@ -31,35 +35,92 @@ import com.pspdfkit.listeners.OnPreparePopupToolbarListener
...
 
-    private lateinit var pdfFragment: PdfFragment
-    private val psPdfKitListener = PsPdfKitListener()
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        // We don't support fragment restoration for the PdfFragment, as we want to recreate a fresh
+        // instance in [reset]. To prevent restoring (and crashing) the PdfFragment without a
+        // document source, we remove it from the fragment manager.
+        (childFragmentManager.findFragmentByTag(pdfFragmentTag) as? PdfFragment)
```
mickael-menu, line 117: Thankfully, PSPDFKit's `PdfFragment` will crash when creating its view and not when creating the fragment, so we have time to remove it from the fragment manager.

readium/navigator/src/main/java/org/readium/r2/navigator/input/InputListener.kt
```diff
@@ -9,17 +8,17 @@ public interface InputListener {
      * Called when the user tapped the content, but nothing handled the event internally (eg.
      * by following an internal link).
      */
-    public fun onTap(navigator: VisualNavigator, event: TapEvent): Boolean = false
```
mickael-menu, line 12: I simplified `InputListener` to not depend on a `Navigator` instance. This way it could be forwarded to a lower level component, such as a `PdfDocumentFragment`.

readium/navigator/src/main/java/org/readium/r2/navigator/pdf/PdfNavigatorFactory.kt
```diff
@@ -38,13 +38,15 @@ public class PdfNavigatorFactory<S : Configurable.Settings, P : Configurable.Pre
     public fun createFragmentFactory(
         initialLocator: Locator? = null,
         initialPreferences: P? = null,
-        listener: PdfNavigatorFragment.Listener? = null
+        listener: PdfNavigatorFragment.Listener? = null,
+        documentFragmentListener: L? = null
```
mickael-menu, line 42: I removed the `PdfEngineProvider`'s `appListener` but added a `documentFragmentListener` to handle configuration of the PDF fragments and intercepting `Resource.Exception`.

qnga: `onResourceLoadFailed` is common between both ` PdfDocumentFragment.Listener` implementations, though it is not mentioned in the interface. I feel like it could be forwarded to `PdfNavigatorFragment.Listener`. What was wrong with having the implementation-specific listener passed to the provider constructor?

readium/adapters/pdfium/pdfium-document/src/main/java/org/readium/adapters/pdfium/document/PdfiumDocument.kt
```diff
@@ -84,24 +86,28 @@ public class PdfiumDocumentFactory(context: Context) : PdfDocumentFactory<Pdfium
 
     private val core by lazy { PdfiumCore(context.applicationContext) }
 
-    override suspend fun open(file: File, password: String?): PdfiumDocument =
-        core.fromFile(file, password)
+    override suspend fun open(file: File, password: String?): ResourceTry<PdfiumDocument> =
```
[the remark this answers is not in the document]
mickael-menu: Doesn't look like we do.

qnga (approved):

PR #400, opened 2023-09-28
Prevent out of memory errors when sniffing a media type

mickael-menu:
This restores [this fix](https://github.com/readium/r2-shared-kotlin/commit/dd0780268539aa2376f89c84a7ab61fed83d62a5#diff-be198afd4545d50e95947243a040c05085a51ae67be0a16f45851106955df941) which was lost during the media type refactoring.

### DIFF
--- a/readium/shared/src/main/java/org/readium/r2/shared/resource/MediaTypeExt.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/resource/MediaTypeExt.kt
@@ -1,6 +1,7 @@
 package org.readium.r2.shared.resource
 
 import org.readium.r2.shared.util.Url
+import org.readium.r2.shared.util.getOrDefault
 import org.readium.r2.shared.util.mediatype.ContainerMediaTypeSnifferContent
 import org.readium.r2.shared.util.mediatype.ResourceMediaTypeSnifferContent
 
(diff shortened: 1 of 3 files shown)


readium/shared/src/main/java/org/readium/r2/shared/resource/MediaTypeExt.kt
```diff
@@ -21,6 +22,18 @@ public class ContainerMediaTypeSnifferContent(
...
         Url.fromDecodedPath(path)?.let { url ->
-            container.get(url).read(range).getOrNull()
+            container.get(url).safeRead(range)
         }
 }
+
+private suspend fun Resource.safeRead(range: LongRange?): ByteArray? {
+    try {
+        // We only read files smaller than 5MB to avoid an [OutOfMemoryError].
+        if (range == null && length().getOrDefault(0) > 5 * 1000 * 1000) {
```
qnga, line 32: We have a `Resource.Exception.OutOfMemory` so it looks like we're supposed to catch `OutOfMemoryError` in `Resource.read` implementations. Anyway, why do we need this trick rather than catching errors?

mickael-menu: After discussion, we will:
* Remove `Resource.Exception.OutOfMemory` and add a dedicated API to report the OOM issue with PDFium and LCPDF.
* Investigate an alternative to the magic constant of 5MB. For example checking that the file is really a JSON or XML before parsing it, verifying the BOM UTF-8, etc.

In the meantime I'm merging this one to address the crash.

PR #401, opened 2023-09-28
Fix various crashes

mickael-menu:


readium/navigator/src/main/java/org/readium/r2/navigator/R2WebView.kt
```diff
@@ -1092,3 +1092,15 @@ internal class R2WebView(context: Context, attrs: AttributeSet) : R2BasicWebView
         }
     }
 }
+
+/**
+ * May crash with java.lang.IllegalArgumentException: pointerIndex out of range
+ */
+private fun MotionEvent.safeGetX(pointerIndex: Int): Float =
+    try { getX(pointerIndex) } catch (e: IllegalArgumentException) { 0F }
```
qnga, line 1100: Is falling back to 0 the right behaviour every time? I feel like we might be able to do something better.

qnga (approved):

readium/navigator/src/main/java/org/readium/r2/navigator/R2WebView.kt
mickael-menu: Yes there is probably a more correct solution, but I'm just trying to avoid crashes in this PR. Since we will deprecate this navigator soonish, I don't want to spend too much time on it.

PR #402, opened 2023-09-30
Remove overscroll between resources in reflowable

qnga:


mickael-menu:
Ideally we would prefer a seamless scroll across resources instead of an overscroll, but IMHO until then the overscroll is still useful as a gesture feedback.

qnga:
As far as I know, overscroll means: there's nothing much on the left/right, please stop swiping. I think that's a misleading feedback because it's definitely not the case between resources in paginated mode.

PR #411, opened 2023-10-05
Support multiple values for HTTP header with same key

qnga:
### Deprecated

 #### Shared

 * `DefaultHttClient.additionalHeaders` is deprecated. Set all the headers when creating a new `HttpRequest`, or modify outgoing requests in `DefaultHttpClient.Callback.onStartRequest()`.

### DIFF
--- a/readium/shared/src/main/java/org/readium/r2/shared/util/http/DefaultHttpClient.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/util/http/DefaultHttpClient.kt
@@ -233,7 +233,7 @@ public class DefaultHttpClient(
             return Try.failure(HttpException.CANCELLED)
         }
 
-        val location = response.valueForHeader("Location")
+        val location = response.header("Location")
             ?: return Try.failure(HttpException(kind = HttpException.Kind.MalformedResponse))
 
         val newRequest = HttpRequest(
@@ -241,7 +241,7 @@ public class DefaultHttpClient(
             body = request.body,
             method = request.method,
             headers = buildMap {
-                response.valuesForHeader("Set-Cookie")
+                response.headers("Set-Cookie")
                     .takeUnless { it.isEmpty() }
                     ?.let { put("Cookie", it) }
             },
--- a/readium/shared/src/main/java/org/readium/r2/shared/util/http/DefaultHttpClient.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/util/http/DefaultHttpClient.kt
@@ -15,6 +15,8 @@ import java.net.URL
 import kotlin.time.Duration
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
+import org.readium.r2.shared.extensions.flatten
+import org.readium.r2.shared.extensions.lowerCaseKeys
 import org.readium.r2.shared.util.Try
 import org.readium.r2.shared.util.flatMap
 import org.readium.r2.shared.util.http.HttpRequest.Method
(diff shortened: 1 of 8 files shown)


readium/shared/src/main/java/org/readium/r2/shared/util/http/DefaultHttpClient.kt
```diff
@@ -39,7 +40,11 @@ import timber.log.Timber
 public class DefaultHttpClient(
     private val mediaTypeRetriever: MediaTypeRetriever,
     private val userAgent: String? = null,
-    private val additionalHeaders: Map<String, String> = mapOf(),
+    @Deprecated(
+        "Pass all headers when building your request or modify it in Callback.onStartRequest",
+        level = DeprecationLevel.ERROR
+    )
```
mickael-menu: Unfortunately we can't use `@Deprecated` on a function parameter. I added a second constructor to deprecate it.

mickael-menu (commented):
I think we should also modify `HttpHeaders.getAll()` to split the values by `,` here: https://github.com/readium/kotlin-toolkit/blob/1570e80ab0a8f28255e53cabb291b1152f565a5f/readium/shared/src/main/java/org/readium/r2/shared/util/http/HttpHeaders.kt#L46

As the caller is using `getAll()` instead of `get()`, we can infer that the header supports multiple values and so should be split by `,` to expand all the values into a single `List<String>`.

qnga:
That's right. The `get` method could become confusing then though: the caller might call it because he's interested only in the first value, not because he knows that he header can't contain multiple values. So `get` would require him to split values while `getAll` would not. What do you think?

qnga:
Actually, public methods are `valueForHeader` and `valuesForHeader`. I wonder if the best solution wouldn't be that `valueForHeader` must be used for single-value headers and `valuesForHeader` for multiple-value headers. Alternatively, we could leave all splitting to callers as we do now in my PR.

mickael-menu:
> I wonder if the best solution wouldn't be that valueForHeader must be used for single-value headers and valuesForHeader for multiple-value headers.

That's what I had in mind. But we could also have `valueForHeader` join all the field values with a comma if there are several. If the caller is interested in the first value, they should use `valuesForHeader().first`.

Unfortunately we can't preemptively split the values as some single-value headers are allowed to contain a comma, e.g.

```http
Expires: Wed, 21 Oct 2015 07:28:00 GMT
```

qnga:
Unfortunately, even single-value headers can have multiple occurrences. For instance, I'm afraid that none of both methods would be suitable for reading a header such as `Set-Cookie`.
`valueForHeader` would return either only the first occurrence or all the values concatenated in a form difficult to split back. On the other hand, `valuesForHeader` would dangerously split values at commas.

OkHttp's strategy seems the wisest choice to me.

mickael-menu:
You're right, HTTP headers are a mess! Which strategy OkHttp uses?

qnga:
> It is the caller’s responsibility to detect and split on commas if their field permits multiple values.

 Actually they still offer a helper to get the raw value of the _last_ header with a given name.

In our case, that would mean offering `headers(name: String): List<String>` and `header(name: String): String?` (I feel like `value` could imply true single value). This way, a caller interested in only one value of a header not allowing comma-separated list would only have to call `header(name: String)`.

mickael-menu:
Sounds good. With `header()` returning the last value for the name, instead of the first?

qnga:
Yes! I updated the interface.

mickael-menu (approved):

PR #412, opened 2023-10-06
Relax rules when parsing a RWPM

mickael-menu:
We have incorrect RWP manifest and packages in the wild:

* missing a `self` link in a manifest,
* with percent-decoded paths instead of valid URL in `Link` objects' `href`.

This PR relaxes the checking rules to avoid breaking backward compatibility.

Note that there is an ambiguity with percent-decoded paths. We favor valid `href` first (valid percent-encoded URL) and fallback on a percent-decoded path only if it fails.


readium/shared/src/main/java/org/readium/r2/shared/publication/Href.kt
```diff
@@ -22,19 +22,18 @@ import timber.log.Timber
...
-                }
-
-            return url?.let { Href(it) }
-        }
+        /**
+         * Creates an [Href] from an URL template (RFC 6570).
+         */
+        @Suppress("RedundantNullableReturnType") // In case we validate the URI template at some point.
+        public fun fromTemplate(href: String): Href? =
+            Href(TemplatedUrl(href))
```
qnga, line 36: What's the rationale? The previous version looks more consistent and convenient to me.

mickael-menu: The previous API was not convenient to do [the fallback with percent-decoded paths](https://github.com/readium/kotlin-toolkit/pull/412/files#diff-e19d5fe6f81c3453c4c498c325f1c85f2302c159eddc8b902fb4da18509b1e55R203-R219
).

We could keep it but it would be used only in [the LCP `Link` model](https://github.com/readium/kotlin-toolkit/pull/412/files#diff-5e7c723489ee0813620c4ae4548960f4f057aab3b9085b11924d4b2c5e003937R41-R47). IMHO it's a bit weird to have both the separate constructors + a convenience constructor which just relocate the condition.

qnga: It used to be largely used on literals in tests.

I'm not sure what you would find a bit weird, but I can propose the following API which should fit every need:
- `invoke(href: String, templated: Boolean = false`)
- `invoke(url: Url)`

qnga (approved):

PR #432, opened 2024-01-09
Small changes for v3 and guide to open a publication

qnga:


readium/shared/src/main/java/org/readium/r2/shared/util/mediatype/MediaTypeHints.kt
```diff
@@ -9,38 +9,40 @@ package org.readium.r2.shared.util.mediatype
...
 /**
- * Bundle of media type and file extension hints for the [MediaTypeSniffer].
+ * Bundle of media type and file extension hints for the MediaTypeSniffer.
  */
+@Suppress("UNUSED_PARAMETER")
+@Deprecated(
+    "Use FormatHints instead.",
+    level = DeprecationLevel.ERROR,
+    replaceWith = ReplaceWith("FormatHints")
+)
```
mickael-menu, line 19: This was never released so it can be removed.

PR #441, opened 2024-01-12
Update the documentation for Readium 3.0.0-alpha.1

mickael-menu:
* Updated the [migration guide](https://github.com/readium/kotlin-toolkit/blob/migration/docs/migration-guide.md) and other documentation.
* Added missing deprecation notices.
* Renamed `State.Error` into `State.Failure` to avoid the conflict/confusion with `Error`.


docs/guides/open-publication.md
```diff
@@ -1,33 +1,58 @@
...
+// or from a raw URL string
+val url = AbsoluteUrl("https://domain/book.epub")
+
+val asset = assetRetriever.retrieve(url)
+    .getOrElse {  /* Failed to retrieve the Asset */  }
+```
 
-The easiest way to get a `PublicationParser` is to use the `DefaultPublicationParser` class. As to
-`ContentProtection`s, you can get one to support LCP publications through your `LcpService` if you have one.
+The `AssetRetriever` will sniff the format of the asset, which you can store in your bookshelf database to speed up the process next time you retrieve the `Asset`. This will improve performance, especially with HTTP URL schemes.
```
qnga, line 39: Confusing. The `Format` is not what's supposed to be stored (it's not even deserializable).

docs/migration-guide.md
```diff
@@ -2,36 +2,59 @@
...
+## 3.0.0-alpha.1
 
-## 2.4.0
+First of all, upgrade to version 2.4.0 and resolve any deprecation notices. This will help you avoid troubles, as the APIs that were deprecated in version 2.x have been removed in version 3.0.
 
-### Maven Central
+### Minimum requirements
 
-Readium is now distributed with [Maven Central](https://search.maven.org/search?q=g:org.readium.kotlin-toolkit). You must update your Gradle configuration.
+Readium 3.0 requires Kotlin 1.9.0 and Gradle 8.0.0. You should start by updating these dependencies in your application.
```
qnga, line 11: Is this true if you use Readium as a regular external dependency?

docs/migration-guide.md
```diff
@@ -2,36 +2,59 @@
...
+    .getOrElse { /* Failed to access the asset or recognize its format */ }
+    .mediaType
 ```
 
-### Migration of HREFs and Locators (bookmarks, annotations, etc.)
+### HREFs
+
+#### `Link.href` and `Locator.href` are not strings anymore
+
+`Link.href` and `Locator.href` are now respectively `Href` and `Url` objects. If you still need the string value, you can call `toString()`, but you may find the `Url` objects more useful in practice.
```
qnga, line 55: How do we get an `Url` from `Link.href`?

docs/migration-guide.md
```diff
@@ -80,29 +103,235 @@ val MIGRATION_HREF = object : Migration(1, 2) {
...
+    .getOrElse { throw ErrorException(it) }
+```
+`UserException` is also deprecated. The application now needs to provide localized error messages for toolkit errors.
+
+### Navigator
+
+#### Click on external links
+
+Clicking on external links is no longer managed by the navigator. To open the link yourself, override `HyperlinkNavigator.Listener.onExternalLinkActivated`, for example:
+
```
qnga, line 123: By the EPUB navigator? I'm not so sure about PDF.

docs/migration-guide.md
```diff
@@ -80,29 +103,235 @@ val MIGRATION_HREF = object : Migration(1, 2) {
...
+        .let { LicenseDocument.fromBytes(it) }
+        .getOrElse { /* Failed to parse a valid LCP License Document from the the raw bytes */ }
+
+    // 4. Download the publication using the `LcpPublicationRetriever`.
+    // The returned `requestId` can be used to cancel an on-going download, or to resume a download
+    // with `LcpPublicationRetriever.register()`, if it was downloaded in the background.
+    val requestId = lcpService.publicationRetriever()
+        .retrieve(license, listener = object : LcpPublicationRetriever.Listener {
+            override fun onAcquisitionCompleted(
+                requestId: LcpPublicationRetriever.RequestId,
```
qnga, line 211: It should be cleared that the callbacks can be called after an app restart.

docs/migration-guide.md
```diff
@@ -137,94 +366,41 @@ To avoid conflicts when merging your app resources, all resources declared in th
...
-})
+}
 ```
 
-### New navigator interfaces
-
-An important part of `VisualNavigator` was moved to `DirectionalNavigator`. A new `HyperlinkNavigator`
-interface implemented by the `EpubFragmentNavigator` provides the ability to intercept clicks on
-links to both internal and external URLs. Opening external URLs is no longer handled by the navigator,
-that's something you should do by yourself in the way you like.
```
qnga, line 176: Is this still elsewhere?

readium/navigators/media/tts/src/main/java/org/readium/navigator/media/tts/TtsNavigator.kt
```diff
@@ -183,13 +183,13 @@ public class TtsNavigator<S : TtsEngine.Settings, P : TtsEngine.Preferences<P>,
...
+            is TtsPlayer.State.Failure -> this.toError()
         }
 
-    private fun TtsPlayer.State.Error.toError(): State.Error =
+    private fun TtsPlayer.State.Failure.toError(): State.Failure =
         when (this) {
-            is TtsPlayer.State.Error.ContentError -> State.Error(Error.ContentError(error))
-            is TtsPlayer.State.Error.EngineError<*> -> State.Error(Error.EngineError(error))
+            is TtsPlayer.State.Failure.ContentFailure -> State.Failure(Error.ContentError(error))
+            is TtsPlayer.State.Failure.EngineFailure<*> -> State.Failure(Error.EngineError(error))
```
qnga, line 192: `Failure.Content` and `Failure.Engine` might be more consistent with the rest?

readium/shared/src/main/java/org/readium/r2/shared/util/Try.kt
```diff
@@ -113,6 +113,20 @@ public fun <S, F : Throwable> Try<S, F>.getOrThrow(): S =
...
+@Suppress("UnusedReceiverParameter")
+@Deprecated(
+    "An `Error` is not a throwable object. Refactor or wrap in an `ErrorException`.",
+    ReplaceWith("getOrElse { throw ErrorException(it) }"),
+    DeprecationLevel.ERROR
+)
+@JvmName("getOrThrowError")
+public fun <S, F : Error> Try<S, F>.getOrThrow(): S =
+    throw NotImplementedError()
+
```
qnga, line 129: I doubt this could actually appear to API consumers because there were no such things as `Error`s before (if I remember correctly).

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
```diff
@@ -359,7 +359,9 @@ private fun String.isValidUrl(): Boolean =
 @JvmInline
 public value class FileExtension(
     public val value: String
-)
+) {
+    override fun toString(): String = value
+}
```
qnga, line 364: Isn't this implicit with value classes?

docs/migration-guide.md
mickael-menu: With `link.url()`, which is an alias of `link.href.resolve()`.

docs/migration-guide.md
mickael-menu: There are `@Deprecated` notices that will explain the moves. For explaining the new interfaces, I think we can do that in the Navigator user guide.

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
mickael-menu: Unfortunately no, it's similar to data classes: `FileExtension(value = "epub")`.

I used this to simplify stuff like: "filename.$fileExtension".

readium/shared/src/main/java/org/readium/r2/shared/util/Try.kt
mickael-menu: We have a few APIs that use to return `Try<X, Exception>` that now return `Try<X, Error>`. This will suggest migrating to a proper handling of errors, or to rethrow a `ErrorException`.

readium/navigators/media/tts/src/main/java/org/readium/navigator/media/tts/TtsNavigator.kt
mickael-menu: Yes, good idea.

docs/migration-guide.md
qnga: Haha, _I_ know. I think the reader might have the same question though.

readium/shared/src/main/java/org/readium/r2/shared/util/Try.kt
qnga: Got it.

docs/migration-guide.md
mickael-menu: Yes, with Pdfium it's not supported for sure. PSPDFKit handles external links, I'm not sure we can intercept it.

docs/migration-guide.md
mickael-menu: I briefly mentioned it in the comment above, but I think that should be part of the download manager user guide and/or the LCP one (I intend to write one soon). For the migration, I expect people will begin with the `ForegroundDownloadManager`.

docs/migration-guide.md
mickael-menu: Yes for the Kotlin version, [see this Slack discussion](https://readium.slack.com/archives/C703MSTQU/p1623427325075300).

For Gradle 8.0, I guess it's only required when not using Maven Central, although I'm not 100% sure.

qnga (approved):

PR #444, opened 2024-01-15
Display error details in the Test App

mickael-menu:
Add a "Details" button in the error reporting snackbar to reveal the technical causes of an error. This will help debugging issues in the Test App.

Screenshot_20240115-161709
Screenshot_20240115-161717

### DIFF
--- a/test-app/src/main/java/org/readium/r2/testapp/bookshelf/BookshelfFragment.kt
+++ b/test-app/src/main/java/org/readium/r2/testapp/bookshelf/BookshelfFragment.kt
@@ -153,7 +153,7 @@ class BookshelfFragment : Fragment() {
     private fun handleEvent(event: BookshelfViewModel.Event) {
         when (event) {
             is BookshelfViewModel.Event.OpenPublicationError -> {
-                activity?.let { event.error.toUserError().show(it) }
+                event.error.toUserError().show(requireActivity())
             }
 
             is BookshelfViewModel.Event.LaunchReader -> {
--- a/test-app/src/main/java/org/readium/r2/testapp/bookshelf/BookshelfFragment.kt
+++ b/test-app/src/main/java/org/readium/r2/testapp/bookshelf/BookshelfFragment.kt
@@ -154,28 +154,18 @@ class BookshelfFragment : Fragment() {
     }
 
     private fun handleEvent(event: BookshelfViewModel.Event) {
-        val message =
-            when (event) {
-                is BookshelfViewModel.Event.OpenPublicationError -> {
-                    Timber.e(event.error.toDebugDescription())
-                    (event.error).toUserError().getUserMessage(requireContext())
-                }
-
-                is BookshelfViewModel.Event.LaunchReader -> {
-                    val intent = ReaderActivityContract().createIntent(
-                        requireContext(),
-                        event.arguments
-                    )
-                    startActivity(intent)
-                    null
-                }
+        when (event) {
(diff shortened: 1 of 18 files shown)


test-app/src/main/java/org/readium/r2/testapp/bookshelf/BookshelfFragment.kt
```diff
@@ -154,28 +151,18 @@ class BookshelfFragment : Fragment() {
...
     private fun handleEvent(event: BookshelfViewModel.Event) {
-        val message =
-            when (event) {
-                is BookshelfViewModel.Event.OpenPublicationError -> {
-                    Timber.e(event.error.toDebugDescription())
-                    (event.error).toUserError().getUserMessage(requireContext())
-                }
+        when (event) {
+            is BookshelfViewModel.Event.OpenPublicationError -> {
+                activity?.let { event.error.toUserError().show(it) }
```
qnga, line 156: Can `activity` be null? If so we can fail to deliver feedback. Fortunately, I doubt it.

test-app/src/main/java/org/readium/r2/testapp/drm/DrmManagementFragment.kt
```diff
@@ -122,20 +118,17 @@ class DrmManagementFragment : Fragment() {
...
 }
 
 private fun Date?.toFormattedString() =
     DateTime(this).toString(DateTimeFormat.shortDateTime()).orEmpty()
-
-// FIXME: the toast is drawn behind the navigation bar
-private fun DrmManagementViewModel.DrmError.report(view: View) {
-    Snackbar.make(view, toUserError().getUserMessage(view.context), Snackbar.LENGTH_LONG).show()
-    Timber.w(error.toDebugDescription())
-}
```
qnga, line 141: Is this fixed?

test-app/src/main/java/org/readium/r2/testapp/utils/UserError.kt
```diff
@@ -6,68 +6,84 @@
...
+import org.readium.r2.testapp.utils.extensions.createShareIntent
+import timber.log.Timber
 
 /**
  * An error that can be presented to the user using a localized message.
  */
 class UserError private constructor(
     val content: Content,
-    val cause: UserError?
+    val cause: Error?
```
qnga, line 29: I'd kept this as it was because that allowed to build hierarchical user messages. I'm not sure that's a good idea though. Do you think so too?

qnga (commented):
Good idea!

test-app/src/main/java/org/readium/r2/testapp/drm/DrmManagementFragment.kt
mickael-menu: Most likely, it's using the Activity's content view instead of the fragment. I didn't test it out though.

test-app/src/main/java/org/readium/r2/testapp/utils/UserError.kt
mickael-menu: I don't think we will ever use this in the Test App. But that's an easy fix whenever it's needed: make `UserError` implement `Error`.

qnga: I tried it, that's awkward: `UserError` would need to implement two messages: one for debug without accessing context, and one for user. Never mind.

qnga (approved):

PR #448, opened 2024-01-17
Extract footnote popups from the Epub navigator

### LINKED ISSUE #147: Footnotes behavior difference between iOs and Android
Footnotes could be opened in a popup window, and this is very useful. But unfortunately, I did not find a way to make it work on iOs and Android in the same epub.
I prepared a test epub and tried to keep it as small and simple as possible. There are three types of footnote references in it.
[1] - anchor reference `href="#link_1"` to the same chapter footnote. Opens popup on iOs but on Android it just jumps to footnote.
[2] - full reference `href="../Text/Section0001.xhtml#link_2"` to the same chapter footnote. Jumps to a footnote on iOs. On Android, it jumps to footnote and opens popup at the same time.
[3] - full reference `href="../Text/Section0002.xhtml#link_3"` to the different chapter footnote. Jumps to a footnote on iOs. Opens popup on Android.

[test_book.epub.zip](https://github.com/readium/r2-navigator-kotlin/files/5423890/test_book.epub.zip)

### LINKED ISSUE #446: UserPreferencesBottomSheetDialogFragment does not survive configuration changes
### Describe the bug

Test app crash when device is rotated while the user settings bottom sheet dialog is opened.

### How to reproduce?

1. Open book
2. Open user settings bottom sheet
3. Rotate the device

### Readium version

2.4.0

### Android API version

33

### Additional context

java.lang.RuntimeException: Unable to start activity ComponentInfo{org.readium.r2reader/org.readium.r2.testapp.reader.ReaderActivity}: androidx.fragment.app.Fragment$InstantiationException: Unable to instantiate fragment org.readium.r2.testapp.reader.preferences.UserPreferencesBottomSheetDialogFragment: could not find Fragment constructor

qnga:
Activation of links to footnotes now triggers `HyperlinkNavigator.shouldFollowInternalLink` with a `FootnoteContext`. Apps can show the footnote content in a popup, as the navigator used to do in a inconsistent way.

Closes #147 
Closes #446


readium/navigator/src/main/java/org/readium/r2/navigator/R2BasicWebView.kt
```diff
@@ -128,6 +127,8 @@ internal open class R2BasicWebView(context: Context, attrs: AttributeSet) : WebV
 
     private val uiScope = CoroutineScope(Dispatchers.Main)
 
+    private var urlNotToOverrideLoading: AbsoluteUrl? = null
```
mickael-menu, line 130: Would you mind adding a comment to explain why this is needed?

readium/navigator/src/main/java/org/readium/r2/navigator/R2BasicWebView.kt
```diff
@@ -596,9 +571,15 @@ internal open class R2BasicWebView(context: Context, attrs: AttributeSet) : WebV
     }
 
     internal fun shouldOverrideUrlLoading(request: WebResourceRequest): Boolean {
-        if (resourceUrl == request.url.toUrl()) return false
```
mickael-menu, line 599: The `if (resourceUrl == request.url.toUrl())` is not needed?

test-app/src/main/java/org/readium/r2/testapp/reader/VisualReaderFragment.kt
```diff
@@ -545,6 +582,56 @@ abstract class VisualReaderFragment : BaseReaderFragment() {
             }
         }
 
+    private fun showFootnote(
+        text: Spanned,
+        point: PointF
+    ) {
+        viewLifecycleOwner.lifecycleScope.launch {
+            viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
```
mickael-menu, line 590: There's no suspending calls here. Did you mean to check for `isAdded`?  I thought the `EventChannel` was handling this already.

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderViewModel.kt
```diff
@@ -272,6 +287,27 @@ class ReaderViewModel(
...
+            is HyperlinkNavigator.FootnoteContext -> {
+                val text =
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                        Html.fromHtml(context.content, Html.FROM_HTML_MODE_COMPACT)
+                    } else {
+                        @Suppress("DEPRECATION")
+                        Html.fromHtml(context.content)
+                    }
+
+                val command = VisualFragmentCommand.ShowPopup(text, context.activationPoint)
```
mickael-menu, line 304: I think there's something fishy with the `activationPoint`, it's not lining up with the link when testing on a device.

readium/navigator/src/main/java/org/readium/r2/navigator/HyperlinkNavigator.kt
```diff
@@ -16,6 +17,19 @@ import org.readium.r2.shared.util.AbsoluteUrl
 @ExperimentalReadiumApi
 public interface HyperlinkNavigator : Navigator {
 
+    @ExperimentalReadiumApi
+    public sealed interface LinkContext {
+        public val referrer: String
+        public val activationPoint: PointF
```
mickael-menu, line 23: What happens if we activate a link using a keyboard interaction? Maybe we should have an `activationEvent: ActivationEvent?` with `{ KeyboardEvent, TapEvent(PointF) }`?

qnga: That's not possible at the moment, is it? Or at least, it would not be intercepted. Actually, I may just have picked up a wrong name. I believe the point of interest is the location of the noteref, because the popup should be anchored there.

readium/navigator/src/main/java/org/readium/r2/navigator/R2BasicWebView.kt
qnga: I don't think we should give links to the same page a special treatment. Apps might want to intercept them too.

test-app/src/main/java/org/readium/r2/testapp/reader/VisualReaderFragment.kt
qnga: `repeatOnLifecycle` is a suspending call... This is a way to handle configuration changes.

test-app/src/main/java/org/readium/r2/testapp/reader/ReaderViewModel.kt
qnga: I can restore the previous behaviour until we design a pretty scrollable footnote popup.

readium/navigator/src/main/java/org/readium/r2/navigator/R2BasicWebView.kt
mickael-menu: If they intercept it you would end up with a blank page. But for simplicity's sake, that's fine by me.

test-app/src/main/java/org/readium/r2/testapp/reader/VisualReaderFragment.kt
qnga: That's true, I guess closing the popup should cancel the job.

qnga: I've tried some fixes, but I can't reason about lifecycles in `VisualReaderFragment`, so it's presumably broken. Fixing it would require significantly more work (probably a rewrite).

mickael-menu (approved):

PR #453, opened 2024-01-24
Use canonical EPUB HREFs

mickael-menu:
We encountered an issue in production when an LCP-protected EPUB uses HREFs with different encoding in the OPF file and the `encryption.xml` one.

For example with the entry `OEBPS/Epub_LaDeuxièmeVie-1.xhtml`, the OPF contains:

```xml
<item id="Epub_LaDeuxie-meVie-1" href="Epub_LaDeuxie%CC%80meVie-1.xhtml" media-type="application/xhtml+xml" />
```

but `META-INF/encryption.xml`:

```xml
<CipherData xmlns="http://www.w3.org/2001/04/xmlenc#">
  <CipherReference xmlns="http://www.w3.org/2001/04/xmlenc#" URI="OEBPS/Epub_LaDeuxièmeVie-1.xhtml"></CipherReference>
</CipherData>
```

This prevented the `encrypted` property being filled as the HREFs didn't match:
https://github.com/readium/kotlin-toolkit/blob/c9a09acc629713eba2fd83878464fe126407346a/readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/ResourceAdapter.kt#L100-L102

As a workaround, we can refuse an URL from containing non ASCII characters like `è` ([as it should](https://url.spec.whatwg.org/#percent-encoded-bytes)).

(@qnga There's a sample LCPL in APP-521)


readium/shared/src/main/java/org/readium/r2/shared/publication/epub/EpubEncryptionParser.kt
```diff
@@ -29,7 +29,7 @@ public object EpubEncryptionParser {
     private fun parseEncryptedData(node: ElementNode): Pair<Url, Encryption>? {
         val resourceURI = node.getFirst("CipherData", Namespaces.ENC)
             ?.getFirst("CipherReference", Namespaces.ENC)?.getAttr("URI")
-            ?.let { Url(it) ?: Url.fromDecodedPath(it) }
```
qnga, line 32: I don't really understand why that was not enough. It should output a canonical path:
- if he input is encoded, then `Url.path` decodes it.
- it the input is decoded, then `Url.fromDecodedPath` encodes it, and then `Url.path` decodes it back.

mickael-menu: That would work if we used a `String` path in `LcpDecryptor.encryptionData`, but we use a `Url` as it might not be a `path` (remote resources). Then the problem is that you can have the same path encoded differently in the URL, but it still represent the same resource:

* `OEBPS/Epub_LaDeuxi%C3%A8meVie-1.xhtml`
* `OEBPS/Epub_LaDeuxie%CC%80meVie-1.xhtml`

But now that I'm saying this, with LCP we always decrypt packaged (relative) resources, so maybe we could actually use a `String` path in `encryptionData`?

mickael-menu: I was mistaken with my last comment, the issue is that `é` is valid in an URI, but can still be percent-encoded as `e%CC%80`. So we need the normalization step to have a canonical URI for the comparison.

qnga (approved):

PR #454, opened 2024-01-30
Implement sniffing of CSS and JS files

io7m:
This adds a couple of media type sniffers that can guess CSS and JS files from their content types and/or file extensions.

I added them as the lowest priority they could be.


io7m:
Wrong branch!

PR #455, opened 2024-01-30
Implement sniffing of CSS and JS files

io7m:
This adds a couple of media type sniffers that can guess CSS and JS files from their content types and/or file extensions.


qnga (commented):
Thank you! Your unit tests are quite weird. It looks like you test idempotence of your sniffers and only that.

io7m:
There didn't appear to be any other tests for the sniffers that I could find. I wasn't sure how to test for "the sniffer guesses correctly when given a file extension".

Is there an example you can point me at?

qnga:
For what I think are historical reasons, we basically only have tests for the behavior of all sniffers together. There : https://github.com/readium/kotlin-toolkit/blob/develop/readium/shared/src/test/java/org/readium/r2/shared/util/asset/AssetSnifferTest.kt

qnga (approved):
Looks good to me.

PR #456, opened 2024-01-31
Cleanup Gradle and Version files, fix Dokka

stevenzeck:
- Fix for Dokka error when running the docs build, see https://github.com/Kotlin/dokka/issues/2954#issuecomment-1694714170.
- Apply the Dokka plugin to all subprojects excepts for test-app
- Remove the old clean gradle task, it does the same thing as normal clean so it's pointless.
- Update several more libraries
- Restructure dependencies and bundles in the versions and gradle files. Note that coroutines-core was never needed, coroutines-android alone sufficed.


mickael-menu (approved):
Thank you @stevenzeck

PR #457, opened 2024-01-31
Fix DownloadManager to better support process killing

qnga:


test-app/src/main/java/org/readium/r2/testapp/domain/Bookshelf.kt
```diff
@@ -110,19 +112,21 @@ class Bookshelf(
...
     }
 
     private suspend fun addBook(
         url: AbsoluteUrl,
+        format: Format? = null,
         coverUrl: AbsoluteUrl? = null
     ): Try<Unit, ImportError> {
         val asset =
-            assetRetriever.retrieve(url)
+            assetRetriever.retrieve(url, FormatHints(format?.mediaType))
```
mickael-menu, line 129: Why don't you use the `format` directly here?

test-app/src/main/java/org/readium/r2/testapp/Application.kt
```diff
@@ -79,14 +80,22 @@ class Application : android.app.Application() {
...
                 createPublicationRetriever = { listener ->
+                    val publicationImporter = PublicationImporter(
+                        listener = listener,
+                        storageDir = storageDir,
+                        assetRetriever = readium.assetRetriever
+                    )
+
                     PublicationRetriever(
                         listener = listener,
+                        publicationImporter = publicationImporter,
```
mickael-menu, line 91: I think you can simplify a bit the `PublicationImporter` and having it only as a dependency of the `LocalPublicationRetriever`.
As in `PublicationImporter`, it's only used in the `LocalPublicationRetriever`'s listener.

readium/shared/src/main/java/org/readium/r2/shared/util/downloads/DownloadManager.kt
```diff
@@ -85,11 +92,11 @@ public interface DownloadManager {
 
     /**
      * Submits a new request to this [DownloadManager]. The given [listener] will automatically be
-     * registered.
+     * registered. Requests already submitted will be ignored.
      *
      * Returns the ID of the download request, which can be used to cancel it.
      */
-    public fun submit(request: Request, listener: Listener): RequestId
+    public fun submit(request: Request, listener: Listener)
```
mickael-menu, line 99: It could be worth mentioning that we *must* call `remove()` when we're done in the `Listener`, no matter the result.

readium/shared/src/main/java/org/readium/r2/shared/util/downloads/DownloadManager.kt
```diff
@@ -25,10 +26,16 @@ import org.readium.r2.shared.util.mediatype.MediaType
  */
 public interface DownloadManager {
 
-    public data class Request(
-        val url: AbsoluteUrl,
-        val headers: Map<String, List<String>> = emptyMap()
-    )
+    public class Request private constructor(
```
mickael-menu, line 29: Why prevent giving a custom ID?

readium/shared/src/main/java/org/readium/r2/shared/util/downloads/android/AndroidDownloadManager.kt
```diff
@@ -98,6 +93,11 @@ public class AndroidDownloadManager internal constructor(
     private val listeners: MutableMap<DownloadManager.RequestId, MutableList<DownloadManager.Listener>> =
         mutableMapOf()
 
+    /*
+     * If the process is killed at a bad time, listeners for a given [requestId] can be called
```
mickael-menu, line 97: Should we mention this in the `DownloadManager` interface? Otherwise, it could be easily overlooked.

readium/lcp/src/main/java/org/readium/r2/lcp/LcpPublicationRetriever.kt
```diff
@@ -183,27 +201,23 @@ public class LcpPublicationRetriever(
             download: DownloadManager.Download
         ) {
             coroutineScope.launch {
-                val lcpRequestId = RequestId(requestId.value)
+                val data = downloadsRepository.getDownload(requestId)
+                    ?: return@launch // Repository is corrupted.
```
mickael-menu, line 205: We don't need to report this error to the listener?

(same for other occurrences below)

test-app/src/main/java/org/readium/r2/testapp/domain/Bookshelf.kt
qnga: Because we don't always have it, and the `AssetRetriever` doesn't accept an optional format. If I remember correctly, we even considered removing its method taking a `Format`.

readium/shared/src/main/java/org/readium/r2/shared/util/downloads/DownloadManager.kt
qnga: I am pretty glad with this solution actually. It nicely prevents ID conflicts between the toolkit and the app. What would be the point of custom IDs?

readium/shared/src/main/java/org/readium/r2/shared/util/downloads/android/AndroidDownloadManager.kt
qnga: I don't remember the point of saying this. It looks like a clarifying note to myself. I think the contract is now quite clear: while the download is not removed, every listener for a request (initially present or added later) will be called once (excepted for progress of course). Not sure if and where we should explain that.

readium/lcp/src/main/java/org/readium/r2/lcp/LcpPublicationRetriever.kt
qnga: I think it could be interesting to inform the user in that case. But to do this, we would need to add a member to `LcpError` hierarchy or design a specific one, which I was not willing to do.

test-app/src/main/java/org/readium/r2/testapp/Application.kt
mickael-menu: You can try this patch I made for what I had in mind: [simplify.diff.txt](https://github.com/readium/kotlin-toolkit/files/14169264/simplify.diff.txt). I think it could be simplified further actually, I'm not sure why we need the `PublicationImporter` instead of a private function in `LocalPublicationRetriever` here.

I couldn't check if I broke something though, because both OPDS and LCP downloads were already broken in the PR.

qnga: What's wrong with OPDS and LCP? They perfectly work on my device.

mickael-menu: I get these errors.

Screenshot_1707155725
Screenshot_1707155738

mickael-menu: I just double-checked, and it works from `develop`.

readium/lcp/src/main/java/org/readium/r2/lcp/LcpPublicationRetriever.kt
mickael-menu: Worst case, we have `LcpError.Runtime` and `LcpError.Unknown`. I'm concerned that the app would get stuck on this download if no error is reported there, what would happen?

test-app/src/main/java/org/readium/r2/testapp/Application.kt
qnga: It's wrong because the contract is that `PublicatinRetriever` always return a file in the the library dir. That's exactly the purpose of `PublicationImporter`.

readium/lcp/src/main/java/org/readium/r2/lcp/LcpPublicationRetriever.kt
qnga: Oh, no, I know why I did that. You have no choice because you can't get the request ID to pick up the listeners to call.

mickael-menu: Ha yeah, makes sense.

PR #473, opened 2024-02-19
Add LcpService.injectLicenseDocument

qnga:
`LcpService.injectLicenseDocument` enables apps to download publications with a custom engine (possibly working in background) and then manually inject the license document into the publication file.


mickael-menu (approved):
LGTM, thanks

PR #476, opened 2024-02-26
Prevent infinite recursion in format sniffing

qnga:
FormatSniffers were able to indefinitely correct each other by returning a different media type. Now:
- `sniffContent` can correct `sniffHints`
- `sniffContent` cannot correct a previous call to `sniffContent`, meaning it must add a specification to be able to change the media type or the file extension.


mickael-menu (approved):
LGTM, it also simplifies nicely the sniffers 👌

PR #484, opened 2024-04-04
Fix various crashes

mickael-menu:
Fix various crashes reported on Firebase Crashlytics.

* [Fix race condition in `TtsPlayer.restartUtterance()`](https://github.com/readium/kotlin-toolkit/commit/4653ec1151c1c1e0310f9c527ed4eefc181bd72d)
* [Fix crash when opening a PDF, when the `ResourceDataProvider` is not used while initializing](https://github.com/readium/kotlin-toolkit/commit/b68a60f9fd45114a9d814a199c9493d839665f4a) 
* [Fix crash in `EpubDeobfuscator`, if the obfuscation key is incorrect](https://github.com/readium/kotlin-toolkit/commit/6d58aac65a20b31098c0c7e1d233de7a2056bc0f)
* [Fix crash when attempting to open a `content://` asset without the permission](https://github.com/readium/kotlin-toolkit/commit/b97528c5498d1c08a17fb252a1f1ae628309b412) 
    * In this case, it will fail with a `java.lang.SecurityException: ... has no access to content://media/external/file/...`
* [Fix a crash when deflating an incorrect ZIP byte array](https://github.com/readium/kotlin-toolkit/pull/484/commits/bc97bbd1dc7881ab7278f3a93a17d0919f9ec202)
* [Fix crash when an LCP resource has an invalid padding length](https://github.com/readium/kotlin-toolkit/pull/484/commits/b2c065c87c4c9ac09060a530f849faad6d21da1f)


readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/EpubDeobfuscator.kt
```diff
@@ -76,7 +82,17 @@ internal class EpubDeobfuscator(
...
+
+    private fun toHexBytes(string: String): ByteArray? {
+        // Only even-length strings can be converted to an Hex byte array, otherwise it crashes
+        // with StringIndexOutOfBoundsException.
+        if (string.isEmpty() || string.length % 2 != 0) {
+            return null
+        }
+        return string.toHexBytes()
+    }
 }
```
qnga, line 98: In my understanding, only hexadecimal characters can be converted into an hex byte array, so I would rather check in the caller that the parameter looks like an UUID.

qnga (approved):

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/EpubDeobfuscator.kt
mickael-menu: It's an UUID (allegedly?) for Adobe but a SHA-1 digest for IDPF. If we can find an official spec for Adobe I would be okay being more strict. Until then, just checking that we have a non-empty UUID and an even number of characters sounds safer to me.

qnga: I suggest to add this check only for the Adobe case, as a SHA-1 hash looks exactly like an UUID . Look at the implementation of the `toHexBytes` method. How would you encode the letter z? It is not valid in radix 16, so the result of `Character.digit("z", 16)` would be -1, which doesn't make any sense.

qnga: > This is a good point, we can check that it contains only characters from 0 to f.
That's exactly what I mean, I don't care about the ID having "-" or not, we remove them anyway.

qnga: Oh, I was confused by French Wikipedia which claims that an UUID version 5 is a SHA-1 hash. Here is the explanation:
> Version-5 UUIDs are similar, but SHA-1 is used instead of MD5. Since SHA-1 generates 160-bit digests, the digest is truncated to 128 bits before the version and variant bits are replaced.

qnga (approved):

PR #490, opened 2024-04-10
Fix a bunch of thread violations

mickael-menu:
This PR fixes a bunch of thread violations in the toolkit, e.g. when calling to the FS from the main thread.

There are still warnings in the test app itself.

### DIFF
--- a/readium/shared/src/main/java/org/readium/r2/shared/extensions/String.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/extensions/String.kt
@@ -12,30 +12,39 @@ package org.readium.r2.shared.extensions
 import android.net.Uri
 import java.net.URL
 import java.security.MessageDigest
-import java.util.*
-import org.joda.time.DateTime
-import org.joda.time.DateTimeZone
+import java.text.DateFormat
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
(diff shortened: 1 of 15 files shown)


readium/shared/src/main/java/org/readium/r2/shared/extensions/String.kt
```diff
@@ -12,30 +12,39 @@ package org.readium.r2.shared.extensions
...
+    return null
+}
+
+private val iso8601UtcFormats: List<DateFormat> = listOf(
+    "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
+    "yyyy-MM-dd'T'HH:mm:ss'Z'",
+    "yyyy-MM-dd'T'HH:mm:ss",
+    "yyyy-MM-dd'T'HH:mm",
+    "yyyy-MM-dd"
+).map {
```
qnga, line 43: What's the point of this change? A pattern would be easy to miss.

mickael-menu: Two reasons:
* Joda is loading a file from the file system when parsing a date string, which is to be avoided from the main thread. 
* Getting rid of the Joda dependency.

But I'll explore this alternative instead: https://github.com/Kotlin/kotlinx-datetime

qnga (approved):

PR #494, opened 2024-04-15
Fix opening LCP-protected publication in restricted mode

qnga:
Return LcpError.MissingCredentials instead of throwing CancellationException if user cancels authentication


readium/lcp/src/main/java/org/readium/r2/lcp/LcpError.kt
```diff
@@ -21,6 +21,9 @@ public sealed class LcpError(
     override val cause: Error? = null
 ) : Error {
 
+    public object MissingCredentials :
```
mickael-menu, line 24: If it's in `LcpError` and not a shared `ContentProtection` error, I would keep with the LCP terminology and call it `MissingPassphrase`.

mickael-menu (approved):

PR #500, opened 2024-04-26
Revamp format specifications and other changes

qnga:
- Nest `Specification` implementations into the interface
- Switch to the IO thread to read publication cover
- Check publication hash in `LcpService.injectLicense`


readium/lcp/src/main/java/org/readium/r2/lcp/service/LicensesService.kt
```diff
@@ -71,14 +69,20 @@ internal class LicensesService(
         licenseDocument: LicenseDocument,
         publicationFile: File
     ): Try<Unit, LcpError> {
+        licenseDocument.publicationLink.hash
```
mickael-menu, line 72: The `hash` might be missing in the LCP License Document, in which case we don't want to fail the injection.

> Reading Systems should verify the integrity of the downloaded Protected Publication, if a hash is provided.
> https://readium.org/lcp-specs/releases/lcp/latest.html#acquiring-the-publication

readium/streamer/src/main/java/org/readium/r2/streamer/parser/readium/ReadiumWebPubParser.kt
```diff
@@ -85,7 +83,7 @@ public class ReadiumWebPubParser(
         // https://readium.org/lcp-specs/notes/lcp-for-pdf.html
         val readingOrder = manifest.readingOrder
         if (manifest.conformsTo(Publication.Profile.PDF) && formatSpecification.conformsTo(
-                LcpSpecification
+                Specification.Rpf
```
mickael-menu, line 86:
```suggestion
                Specification.Lcp
```

mickael-menu (approved):

PR #506, opened 2024-04-29
Replace `Date` with `Instant`

mickael-menu:
### Removing JVM dependencies

To reduce our depency to the JVM, we no longer use `Date` objects in the toolkit. Instead, we added a custom `Instant` type.

You can still translate from and to a `Date` object with `Instant.fromJavaDate()` and `instant.toJavaDate()`.


readium/lcp/src/main/java/org/readium/r2/lcp/license/model/StatusDocument.kt
```diff
@@ -63,7 +63,7 @@ public class StatusDocument(public val data: ByteArray) {
         public val rawValue: String get() = value
 
         public companion object {
-            public operator fun invoke(value: String): Rel? = values().firstOrNull { it.value == value }
+            public operator fun invoke(value: String): Rel? = entries.firstOrNull { it.value == value }
```
mickael-menu, line 66: That was a deprecation notice in Android Studio.

readium/shared/src/main/java/org/readium/r2/shared/publication/Accessibility.kt
```diff
@@ -59,8 +59,7 @@ public data class Accessibility(
      * Accessibility profile.
      */
     @Parcelize
-    @JvmInline
-    public value class Profile(public val uri: String) : Parcelable {
+    public data class Profile(public val uri: String) : Parcelable {
```
mickael-menu, line 62: Value classes are not supported with KMP.

readium/shared/src/main/java/org/readium/r2/shared/util/Instant.kt
```diff
@@ -0,0 +1,111 @@
...
+import org.readium.r2.shared.InternalReadiumApi
+import org.readium.r2.shared.extensions.tryOrNull
+
+/**
+ * A moment in time.
+ */
+@Parcelize
+@TypeParceler<kotlinx.datetime.Instant, InstantParceler>()
+@Serializable(with = InstantSerializer::class)
+public class Instant private constructor(
```
mickael-menu, line 32: I used a dedicated type instead of using straight `kotlinx.datetime.Instant` to avoid requiring to leak it in our API with `api(libs.kotlinx.datetime)`.

qnga (approved):

PR #521, opened 2024-05-22
Bump dependencies

mickael-menu:
Added explicit opt-ins for `ReadiumInternalApi` because of a bug in AGP, see https://issuetracker.google.com/issues/328105511


README.md
```diff
@@ -10,7 +10,7 @@
 
 | Readium | Android min SDK | Android compile SDK | Kotlin compiler (✻) | Gradle (✻) |
 |---------|-----------------|---------------------|---------------------|------------|
-| 3.0.0   | 21              | 34                  | 1.9.22              | 8.2.2      |
+| 3.0.0   | 21              | 34                  | 1.9.23              | 8.6.0      |
```
qnga, line 13:
```suggestion
| 3.0.0   | 21              | 34                  | 1.9.24              | 8.4.0      |
```

mickael-menu: It's bumped to 8.6.0: https://github.com/readium/kotlin-toolkit/blob/1f42e22018ffb0dca937954a9843024052ee458c/gradle/wrapper/gradle-wrapper.properties

qnga: Not the plugin. Actually I'm not sure which one matters for apps. And the Kotlin version looks still wrong.

stevenzeck: Any reason why not 8.7?

gradle/libs.versions.toml
```diff
@@ -1,73 +1,74 @@
 [versions]
 
-kotlin = "1.9.22"
-agp = "8.2.2"
+kotlin = "1.9.24"
+agp = "8.4.0"
```
stevenzeck, line 4: Mind as well go 8.4.1.

stevenzeck:
Can either the [navigation](https://github.com/readium/kotlin-toolkit/blob/1f42e22018ffb0dca937954a9843024052ee458c/gradle/libs.versions.toml#L172) bundle be removed, or use it in the test app in place of these [two lines](https://github.com/readium/kotlin-toolkit/blob/1f42e22018ffb0dca937954a9843024052ee458c/test-app/build.gradle.kts#L100-L101)?

Can the [test-frameworks](https://github.com/readium/kotlin-toolkit/blob/1f42e22018ffb0dca937954a9843024052ee458c/gradle/libs.versions.toml#L174) bundle also be removed since it's not being used?

README.md
mickael-menu: Ha you're right for the Kotlin compiler. For Gradle it is the Gradle version and not AGP, because this is the one people need to use when integrating the project as a submodule.

> Any reason why not 8.7?

I used what AGP Upgrade Assistant suggested, I'm not sure why 8.7 is not offered. Not yet supported officially maybe?

PR #525, opened 2024-05-28
Update to Kotlin 2.0 and SDK 35

stevenzeck:
Do not merge unless ready to accept Kotlin 2.0. Opening this to make sure the checks pass.

- Upgraded to Kotlin 2.0.x.
- Updated Android SDK to 35.
- KSP2 is not ready as it errors with Room. Leaving it commented out in `gradle.properties`. It won't be ready until Room is updated to 2.7.0 (currently in alpha), and a KSP2 issue with `@TypeConverters`.
- The Compose compiler now ships with Kotlin, instead of relying on keeping that synchronized.
- I used https://github.com/autonomousapps/dependency-analysis-gradle-plugin as a starting point to see which dependencies aren't needed in each module. I did remove all of the `testImplementation` and `androidTestImplementation` ones since they weren't being used. They can always be added back.
- Added a flag to the compiler options for Navigator and LCP modules: `-Xconsistent-data-class-copy-visibility`. See [KT-11914](https://youtrack.jetbrains.com/issue/KT-11914/Confusing-data-class-copy-with-private-constructor).
- Updated Java to 17. AGP 8.7 and/or Android Studio warns about dropping support for Java 8.


stevenzeck:
@mickael-menu This causes the app to crash with the below error when opening an epub. Declaring [R2RTLViewPager](https://github.com/stevenzeck/kotlin-toolkit/blob/a446e9da37b326c98c92d7f5dadce6327d0a43da/readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2RTLViewPager.java#L48) as public fixes it. I don't know why, that old code just goes in a circle.

Possibly related to https://youtrack.jetbrains.com/issue/KT-63050/K1-K2-IllegalAccessError-when-accessing-a-private-field-isnt-detected-at-compile-time. Even though the OP there was using Kotlin 1.6.21 when reporting it.

```
java.lang.IllegalAccessError: Illegal class access: 'org.readium.r2.navigator.epub.EpubNavigatorFragment' attempting to access 'org.readium.r2.navigator.pager.R2RTLViewPager' (declaration of 'org.readium.r2.navigator.epub.EpubNavigatorFragment' appears in /data/app/~~-rjhsDrV6c9hFpC_VngmIw==/org.readium.r2reader-y1pHEqGxpJ6lP_pzngh9Iw==/base.apk!classes4.dex)
at org.readium.r2.navigator.epub.EpubNavigatorFragment.resetResourcePagerAdapter(EpubNavigatorFragment.kt:497)
at org.readium.r2.navigator.epub.EpubNavigatorFragment.resetResourcePager(EpubNavigatorFragment.kt:475)
at org.readium.r2.navigator.epub.EpubNavigatorFragment.onCreateView(EpubNavigatorFragment.kt:414)
```

mickael-menu:
We could set it `public` with an `@InternalReadiumApi` annotation as a workaround. I wonder if this is related to the fact that `R2RTLViewPager` is a Java class. In which case we might have a similar problem when using the streaming ZIP container (e.g. when opening an EPUB from the shared storage).

stevenzeck:
[Fix](https://youtrack.jetbrains.com/issue/KT-55851/K2-reference-to-a-field-from-package-private-class-crashes-in-runtime) will be in 2.0.20.

stevenzeck:
@mickael-menu There is new behavior regarding non-public constructors of data classes and the `copy()` function, see [KT-11914](https://youtrack.jetbrains.com/issue/KT-11914/Confusing-data-class-copy-with-private-constructor). I'm thinking we should annotate these with `@ConsistentCopyVisibility`; I don't think copy should be publicly visible in these cases.

Applicable to:
- [Configuration](https://github.com/readium/kotlin-toolkit/blob/d3f806a34bfa3cfd5c2aeb4ed13dd19dc00b0077/readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt#L142)
- [MutableFontFamilyDeclaration](https://github.com/readium/kotlin-toolkit/blob/d3f806a34bfa3cfd5c2aeb4ed13dd19dc00b0077/readium/navigator/src/main/java/org/readium/r2/navigator/epub/css/FontFamilyDeclaration.kt#L101)
- [MutableFontFaceDeclaration](https://github.com/readium/kotlin-toolkit/blob/d3f806a34bfa3cfd5c2aeb4ed13dd19dc00b0077/readium/navigator/src/main/java/org/readium/r2/navigator/epub/css/FontFamilyDeclaration.kt#L122)
- [Valid](https://github.com/readium/kotlin-toolkit/blob/d3f806a34bfa3cfd5c2aeb4ed13dd19dc00b0077/readium/lcp/src/main/java/org/readium/r2/lcp/license/StateMachine.kt#L83)
- [Invalid](https://github.com/readium/kotlin-toolkit/blob/d3f806a34bfa3cfd5c2aeb4ed13dd19dc00b0077/readium/lcp/src/main/java/org/readium/r2/lcp/license/StateMachine.kt#L89)
- [TransitionTo](https://github.com/readium/kotlin-toolkit/blob/d3f806a34bfa3cfd5c2aeb4ed13dd19dc00b0077/readium/lcp/src/main/java/org/readium/r2/lcp/license/StateMachine.kt#L106)

stevenzeck:
I misread the [phases](https://youtrack.jetbrains.com/issue/KT-11914/Confusing-data-class-copy-with-private-constructor#deprecation-timeline-migration-plan). The @ConsistentCopyVisibility annotation (or -Xconsistent-data-class-copy-visibility flag) are only be necessary until Phase 3. I added the flag to the navigator and LCP gradle files.

qnga:
What about `Edge-to-edge` content being the default with SDK 35 on Android 15? I think this should be tested on an actual Android 15 device as soon as we can.

stevenzeck:
I tested it in the emulator, and yes padding was needed for the status bar. The BottomNavigationView handles that behavior change. `enableEdgeToEdge()` is only if the app should do edge-to-edge on older Android versions.

I did notice an issue in light mode when `enableEdgeToEdge()` is not in there in that the status bar text and icon colors don't contrast with the background. It's not specific to Readium though. I created a bug report.

qnga:
In my understanding, if you don't call `enableEdgeToEdge`, you must apply padding only on Android 15. I didn't test it though.

stevenzeck:
`ViewCompat.setOnApplyWindowInsetsListener` doesn't seem to be called in Android 14 and below if `enableEdgeToEdge` isn't in there. Up to you, but I think it's fine without it. I don't quite understand what they're doing with the status bar color when edge-to-edge is enabled. It doesn't use the dynamic colors.

qnga:
> ViewCompat.setOnApplyWindowInsetsListener doesn't seem to be called in Android 14 and below if enableEdgeToEdge isn't in there. 

Sounds weird. That's something we've been using for a long time at different places. We'll see during tests.

stevenzeck:
Closing in favor of splitting this out.

PR #533, opened 2024-06-11
Improve `Url` equality

mickael-menu:
Instead of comparing URL to their string representation, compare their components with different rules:

* Scheme is case insensitive.
* Paths must be compared percent-decoded.
* Authority, fragment and query parameters must match.

The reason for this change is that some EPUBs encode unreserved characters which breaks locating a Link from an HREF.

For example both of these are valid relative URLs which point to the same resource:

* `Text/Ces_choses_qu'on_laisse.xhtml`
* `Text/Ces_choses_qu%27on_laisse.xhtml`


mickael-menu:
Superseded by https://github.com/readium/kotlin-toolkit/pull/534

PR #534, opened 2024-06-12
Normalize URLs

mickael-menu:
Normalize URLs when comparing them following these rules:

* Scheme is lower case.
* Paths must only percent-encode required characters.
* Relative paths (e.g. `..`) are resolved.

The reason for this change is that some EPUBs encode unreserved characters which breaks locating a Link from an HREF.

For example both of these are valid relative URLs which point to the same resource:

* `Text/Ces_choses_qu'on_laisse.xhtml`
* `Text/Ces_choses_qu%27on_laisse.xhtml`

### DIFF
--- a/readium/shared/src/main/java/org/readium/r2/shared/MediaOverlays.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/MediaOverlays.kt
@@ -22,11 +22,12 @@ public data class MediaOverlays(private val nodes: List<MediaOverlayNode> = list
 
     private fun nodeForFragment(ref: Url?): MediaOverlayNode? = findNode(ref, this.nodes)
 
+    @OptIn(DelicateReadiumApi::class)
     private fun findNode(ref: Url?, inNodes: List<MediaOverlayNode>): MediaOverlayNode? {
         for (node in inNodes) {
             if (node.role.contains("section")) {
                 return findNode(ref, node.children)
-            } else if (ref.isEquivalent(null) || node.text.isEquivalent(ref)) {
+            } else if (ref == null || node.text.isEquivalent(ref)) {
                 return node
             }
         }
@@ -37,6 +38,7 @@ public data class MediaOverlays(private val nodes: List<MediaOverlayNode> = list
 
     private fun nodeAfterFragment(ref: Url?): MediaOverlayNode? = findNextNode(ref, this.nodes).found
 
+    @OptIn(DelicateReadiumApi::class)
     private fun findNextNode(fragment: Url?, inNodes: List<MediaOverlayNode>): NextNodeResult {
         var prevNodeFoundFlag = false
         //  For each node of the current scope...
@@ -56,7 +58,7 @@ public data class MediaOverlays(private val nodes: List<MediaOverlayNode> = list
                     prevNodeFoundFlag = ret.prevFound
                 }
                 //  If the node text refer to filename or that filename is null, return node
-                else if (fragment.isEquivalent(null) || node.text.isEquivalent(fragment)) {
+                else if (fragment == null || node.text.isEquivalent(fragment)) {
                     prevNodeFoundFlag = true
                 }
             }
--- a/readium/shared/src/main/java/org/readium/r2/shared/MediaOverlays.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/MediaOverlays.kt
@@ -11,6 +11,7 @@ package org.readium.r2.shared
 
 import java.io.Serializable
 import org.readium.r2.shared.util.Url
+import org.readium.r2.shared.util.isEquivalent
 
 @InternalReadiumApi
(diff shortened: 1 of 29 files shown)


readium/streamer/src/main/java/org/readium/r2/streamer/parser/audio/AudioParser.kt
```diff
@@ -79,7 +80,7 @@ public class AudioParser(
 
         val readingOrderWithFormat =
             asset.container
-                .mapNotNull { url -> entryFormats[url]?.let { url to it } }
```
qnga, line 82: I think we might want to keep the original URLs here, as we do with other formats.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/image/ImageParser.kt
```diff
@@ -81,7 +82,7 @@ public class ImageParser(
 
         val readingOrderWithFormat =
             asset.container
-                .mapNotNull { url -> entryFormats[url]?.let { url to it } }
+                .mapNotNull { url -> entryFormats.getEquivalent(url)?.let { url to it } }
```
qnga, line 85: Same as for `AudioParser`.

readium/shared/src/test/java/org/readium/r2/shared/util/UrlTest.kt
```diff
@@ -399,4 +399,53 @@ class UrlTest {
...
+
+    @Test
+    fun normalize() {
+        // Scheme is lower case.
+        assertEquals(
+            "http://example.com",
+            Url("HTTP://example.com")!!.normalize().toString()
+        )
+
+        // Path is percent-decoded.
```
qnga, line 411: Legacy comment?

readium/shared/src/main/java/org/readium/r2/shared/MediaOverlays.kt
```diff
@@ -25,7 +26,7 @@ public data class MediaOverlays(private val nodes: List<MediaOverlayNode> = list
         for (node in inNodes) {
             if (node.role.contains("section")) {
                 return findNode(ref, node.children)
-            } else if (ref == null || node.text == ref) {
+            } else if (ref.isEquivalent(null) || node.text.isEquivalent(ref)) {
```
qnga, line 29: IMHO, `== null` remains far clearer.

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
```diff
@@ -173,9 +173,39 @@ public sealed class Url : Parcelable {
...
         uri.toString()
 
+    /**
+     * Returns whether two URLs are strictly equal, by comparing their string representation.
+     *
+     * WARNING: Strict URL comparisons can be a source of bug, if the URLs are not normalized.
+     * In most cases, you should compare using [Url.isEquivalent].
+     */
+    @DelicateReadiumApi
     override fun equals(other: Any?): Boolean {
```
qnga, line 209: Textual `equals` makes perfect sense to me. The only places where I think we _must_ use equivalence are internal data provided referenced by an URL (`LcpDecryptor` and `EpubDeobsfucator`):
- the exchange form of URLs is the one you can find in manifest (or OPF, or Zip entry)
- URLs that apps provide to the toolkit are supposed to come from the readingOrder (what else?), which means they are in the original form.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/audio/AudioParser.kt
mickael-menu: We do, this is used to get the `Format` and it doesn't replace the `Url` of the container.

Although it's not strictly necessary here, the `Url` should be the same instances.

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
mickael-menu: In practice I didn't find any place where it made sense to use a strict comparison, except for when you want to check for `null`. Adding `@DelicateReadiumApi` helped me spot places I missed and is appropriate in my opinion. This is just a warning to make you pause to think if this is really what you want to use.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/audio/AudioParser.kt
qnga: Ok, I read too quick.

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
qnga: That's because you've been using equivalence as a default. I would have used it only when strictly necessary, that is, as I pointed, very rarely. It doesn't matter though.

qnga (approved):

PR #542, opened 2024-06-25
Improvements on ExoPlayer audio navigator

qnga:
- Type Exoplayer errors and report them in testapp
- Optionally use cache in `ExoPlayerEngineProvider`


readium/adapters/exoplayer/audio/src/main/java/org/readium/adapter/exoplayer/audio/DefaultExoPlayerCacheProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+public class DefaultExoPlayerCacheProvider(
+    private val cache: Cache
+) : ExoPlayerCacheProvider {
+
+    @kotlin.OptIn(DelicateReadiumApi::class)
+    override fun getCache(publication: Publication): Cache? =
+        cache.takeUnless { publication.baseUrl == null }
+
+    override fun computeKey(publication: Publication, href: Href): String =
+        href.resolve(publication.baseUrl).toString()
```
mickael-menu, line 24: Should we `normalize()` the URL here?

readium/adapters/exoplayer/audio/src/main/java/org/readium/adapter/exoplayer/audio/ExoPlayerEngineProvider.kt
```diff
@@ -77,6 +82,22 @@ public class ExoPlayerEngineProvider(
...
+    private fun createDataSourceFactory(
+        publication: Publication
+    ): DataSource.Factory {
+        val baseDataSource = ExoPlayerDataSource.Factory(publication)
+        val cache = cacheProvider?.getCache(publication)
+            ?: return baseDataSource
+
+        return CacheDataSource.Factory()
+            .setCache(cache)
+            .setCacheKeyFactory { cacheProvider.computeKey(publication, Href(it.uri.toUrl()!!)) }
```
mickael-menu, line 94: Why pass an `Href` instead of an `Url`, if we have to compute it from a static URL every time?

readium/adapters/exoplayer/audio/src/main/java/org/readium/adapter/exoplayer/audio/DefaultExoPlayerCacheProvider.kt
qnga: I guess it doesn't matter. We're supposed to always pass the same href, so the resulting URL will always be the same.

readium/adapters/exoplayer/audio/src/main/java/org/readium/adapter/exoplayer/audio/ExoPlayerEngineProvider.kt
qnga: I thought the `Href` class conveyed better that what need is a resource identifier. Unfortunately, `Href` can also be templated. I'll change that.

mickael-menu (approved):

PR #545, opened 2024-07-02
Fix applying the PSPDFKit settings on configuration changes

mickael-menu:
There are some cases (not reproducible in the Test App) where the PSPDFKit settings are not applied when the PDF view is recreated.

For example, it can happen that the RTL page binding is not taken into account after a configuration change triggered by forcing an orientation in the parent `Activity` with `setRequestedOrientation()`.


qnga (approved):
Looks good

PR #546, opened 2024-07-16
Allow uses of Try with nullable error types (Try<S, F?>)

qnga:


mickael-menu (approved):

mickael-menu (commented):
I changed my mind. In which case does it make sense to have a nullable failure type instead of a `null` result?

qnga:
To convey the idea that there was a failure we know nothing about. Anyway, there's a mistake in the current code: this kind of types is accepted by the compiler but fails at runtime.

PR #553, opened 2024-08-01
Fix audiobook locators' total progression

mickael-menu:
#### Navigator

* Fixed computing the total progression of audiobook locators.


readium/navigators/media/audio/src/main/java/org/readium/navigator/media/audio/AudioNavigator.kt
```diff
@@ -90,11 +90,12 @@ public class AudioNavigator<S : Configurable.Settings, P : Configurable.Preferen
             val currentItem = readingOrder.items[playback.index]
             val link = requireNotNull(publication.linkWithHref(currentItem.href))
             val item = readingOrder.items[playback.index]
-            val itemStartPosition = readingOrder.items
-                .slice(0 until playback.index)
-                .mapNotNull { it.duration }
-                .takeIf { it.size == readingOrder.items.size }
```
mickael-menu, line 96: This always failed because of the `slice()` above. We only have a portion of the reading order.

PR #562, opened 2024-08-23
Import `android-pdf-viewer` as a JAR instead of an AAR

mickael-menu:
The AAR was not compatible with Maven Central.


qnga (approved):

qnga (commented):
Actually, I guess that the document adapter cannot be used without the navigator adaptor anymore. I think you can create a `common` module containing the JAR in `adapters/pdfium`.

qnga:
Oh, ok, I had no idea.

PR #581, opened 2024-10-25
Speed up access to shared storage

qnga:
WewViews attempt to read all resources at once and by ranges, so range requests reaching a ZIP file are entirely random (for instance range 1 from resource 1, then range 1 from resource 2, then range 2 from resource1, etc.). This PR significantly improves speed (though unfortunately, we still can't reach direct access speed) by
- Reusing the InputStream when reading a range located after the previous read one.
- Reading full ZIP entries in one go and caching them when they are small enough.

In addition, fixes the path that imports from intents go through in the testapp.


mickael-menu (approved):
Looking good, thank you.

PR #587, opened 2024-11-05
Fix value of navigator.overscroll.value.scroll in fxl mode

qnga:


mickael-menu (approved):
Thanks, let's wait a bit before merging this one in case we release `develop` soon.

PR #592, opened 2024-11-22
Fix CBC decryption when originalLength is wrong

qnga:
Fixes `IndexOutOfBoundsException` occurring when the `originalLength` reported in `META-INF/encryption.xml` is incorrect.


mickael-menu (approved):
Thank you for looking into that 🙏

Issue #601, opened 2024-12-30
Kindly help me with audioBook.Lcp Failure when come to the end of file.   


hishampro5:
### Describe the bug

i am developing android app for run  audiobooks.lcp ..etc but the issue only encounter with audiobooks (Lcp) when am trying same file on TestApp show me message that is not supported on test app.

The issue : 
when audio books run it's works fine but when audioBook finsh the error occurs you can see logs below i am not sure if this happen because readium or something else. 
 
 also there are design error when you click on drawer menu twice the back to book  the drawer menu items still appear above of book content. 
 

Error Log : 
UserError: An error occurred in the ExoPlayer engine. ExoPlayerEngine.Error.Engine: An error occurred in the ExoPlayer engine. ExoPlaybackException: Source error androidx.media3.exoplayer.ExoPlayerImplInternal.handleIoException(ExoPlayerImplInternal.java:717) androidx.media3.exoplayer.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:693) Loader.UnexpectedLoaderException: Unexpected IllegalArgumentException: Failed requirement. androidx.media3.exoplayer.upstream.Loader$LoadTask.run(Loader.java:442) java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145) IllegalArgumentException: Failed requirement. org.readium.r2.lcp.LcpDecryptor$CbcLcpResource.lengthFromLastTwoBlocks(LcpDecryptor.kt:150) org.readium.r2.lcp.LcpDecryptor$CbcLcpResource.read(LcpDecryptor.kt:236)

### How to reproduce?

1- open audioBook (LCP) 
2- play the book
3- wait to the end 
4- the error occurs 

### Readium version

3.0.2

### Android API version

14

qnga:
These definitely look like Readium and testapp bugs, thanks for reporting. Your audiobook is probably guilty, but we should handle this case better. Could you confirm that you don't get this log with every LCP audiobook? I don't get it with a random one.

hishampro5:
I'm sorry that this happens with all books of the same type ( Lcp ) 

also we tried on test app and get same issue Test app download from => https://github.com/readium/kotlin-toolkit/releases/download/3.0.2/test-app-3.0.2.apk

[...]

PR #602, opened 2024-12-30
Fix LcpDecryptor and FileZipLicenseContainer

qnga:


mickael-menu:
Could you explain what does this fix and why? Thanks!

qnga:
The diff had spoken for itself. It no longer does though, I basically rewrote everything. I suggest you read the whole `LcpDecryption` file, it's clearer now. I also introduced almost exhaustive tests.

qnga:
Fix #601.

PR #610, opened 2025-01-16
Remove Jetifier

### LINKED ISSUE #609: Remove Jetifier - migrate to AndroidX - Gradle build slowdown
### Describe the bug

Please migrate Readium from Support Library to AndroidX so android.enableJetifier=true is not any more needed. You can check if this is needed using Build Analyzer tool inside Android Studio. Migration guide is available [here](https://developer.android.com/jetpack/androidx/migrate). It should be hopefully easy process.

Why it matters? 
Enabled Jetifier is significatly impaction build times of the Gradle project based on our experiences with other libraries before/after migration to AndroidX.

Thank you very much.

### How to reproduce?

Measure your build time before adding Readium library to your project.
Set android.enableJetifier=true in Gradle project settings.
Measure your build time after.

### Readium version

3.0.3

### Android API version

23 and higher

### Additional context

_No response_

mickael-menu:
### Changed

* Jetifier is not required anymore, you can remove `android.enableJetifier=true` from your `gradle.properties` if you were using Readium as a local clone.

---

* Fixes #609 

---

I addressed this by taking the jetified artifacts in my Gradle cache (running `jetifier-standalone` directly failed) and importing them directly in the project. A downside is that now we have all the native libraries as part of the repository, which adds ~40 MB.


mickael-menu:
This is still a draft because I haven't tested it thoroughly yet.

qnga:
Couldn't you put the JAR into `pdfium-document` libs? `pdium-navigator` depends on it.

mickael-menu:
> Couldn't you put the JAR into `pdfium-document` libs? `pdium-navigator` depends on it.

Not without using `api` instead of `implementation` which would leak the public API of `pdfium-android` into Readium.

qnga:
That's weird, I can see no reference to a PDFium symbol in pdfium-navigator.

mickael-menu:
It uses the symbols indirectly through AndroidPdfViewer APIs.

qnga:
That's what I thought. So I expect that compiling `pdf-navigator` doesn't need PDFium: it only has to be available on the classpath at runtime.

mickael-menu:
It doesn't compile, you can try it out.

qnga:
Okay! I didn't expect gradle to check dependencies of precompiled JAR.

PR #620, opened 2025-02-24
Speed up initialization of `ExoPlayerEngine`

mickael-menu:
We are currently waiting for the first audio resource to buffer before considering the `ExoPlayerEngine` initialized. This can significantly slow down the opening of the audio navigator in case of network latency.

This change will return the audio navigator as soon as we enter the ExoPlayer buffering state. Integrators can then display a spinner to indicate that the audiobook is not yet ready or delay opening the reader to maintain the current behavior.


qnga (approved):
Looks good

PR #624, opened 2025-03-12
feat: add option to disable horizontal swiping when scroll is true.

tm-bookshop:
I'm looking to add the equivalent functionality that was added here: https://github.com/readium/swift-toolkit/pull/531

### DIFF
--- a/readium/navigator/src/main/java/org/readium/r2/navigator/R2BasicWebView.kt
+++ b/readium/navigator/src/main/java/org/readium/r2/navigator/R2BasicWebView.kt
@@ -116,7 +116,7 @@ internal open class R2BasicWebView(context: Context, attrs: AttributeSet) : WebV
     internal var isSelecting = false
 
     val scrollMode: Boolean get() = scrollModeFlow.value
-    var scrollModeDisableSwipePagination: Boolean = false
+    var disablePageTurnsWhileScrolling: Boolean = false
 
     var callback: OnOverScrolledCallback? = null
 
@@ -215,7 +215,7 @@ internal open class R2BasicWebView(context: Context, attrs: AttributeSet) : WebV
 
             when {
                 // If the user is in scrollMode and has disabled swipe pagination, do nothing.
-                scrollMode && this@R2BasicWebView.scrollModeDisableSwipePagination -> {}
+                scrollMode && this@R2BasicWebView.disablePageTurnsWhileScrolling -> {}
 
                 scrollMode ->
                     goRight(jump = true)
@@ -249,7 +249,7 @@ internal open class R2BasicWebView(context: Context, attrs: AttributeSet) : WebV
 
             when {
                 // If the user is in scrollMode and has disabled swipe pagination, do nothing.
-                scrollMode && this@R2BasicWebView.scrollModeDisableSwipePagination -> {}
+                scrollMode && this@R2BasicWebView.disablePageTurnsWhileScrolling -> {}
 
                 scrollMode ->
                     goLeft(jump = true)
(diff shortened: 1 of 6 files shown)


readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorViewModel.kt
```diff
@@ -148,6 +147,8 @@ internal class EpubNavigatorViewModel(
     }
 
     fun onResourceLoaded(webView: R2BasicWebView, link: Link): RunScriptCommand {
+        webView.scrollModeDisableSwipePaginationFlow.value = config.scrollModeDisableSwipePagination
```
tm-bookshop, line 150: This really doesn't feel like the right place to do this but I'm pretty unfamiliar with Kotlin/Android development

tm-bookshop: Done

readium/navigator/src/main/java/org/readium/r2/navigator/R2BasicWebView.kt
```diff
@@ -111,11 +111,13 @@ internal open class R2BasicWebView(context: Context, attrs: AttributeSet) : WebV
     var resourceUrl: AbsoluteUrl? = null
 
     internal val scrollModeFlow = MutableStateFlow(false)
+    internal val scrollModeDisableSwipePaginationFlow = MutableStateFlow(false)
```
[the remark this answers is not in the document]
tm-bookshop: No longer a flow. I agree, it isn't and shouldn't need to be observed anywhere.

qnga:
The name `scrollModeDisableSwipePagination` sounds a bit confusing to me. Pagination? In scroll mode? |scrollModeDisableSwipe` could be enough. It's a bit mysterious but at least not confusing.

tm-bookshop:
> The name `scrollModeDisableSwipePagination` sounds a bit confusing to me. Pagination? In scroll mode?`scrollModeDisableSwipe` could be enough. It's a bit mysterious but at least not confusing.

Fair point @qnga , I thought about swipe initially but thought that was also confusing. Swipes can be up/down or left/right and this only impacts left/right.

The reason I chose "pagination" is because it seemed to fit best with "resource paging" ([via the R2ViewPager](https://github.com/readium/kotlin-toolkit/blob/665d6f81b0e4cecd6d8dbda393cbf788f757fa0e/readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2ViewPager.kt#L19)).  

I'm not really attached to the name though, so happy to with whatever the reviewers land on.

mickael-menu:
@tm-bookshop For the sake of consistency, I suggest using the same name as the one in the Swift toolkit: `Configuration.disablePageTurnsWhileScrolling`.

tm-bookshop:
> @tm-bookshop For the sake of consistency, I suggest using the same name as the one in the Swift toolkit: `Configuration.disablePageTurnsWhileScrolling`.

@mickael-menu , done

mickael-menu (approved):
Thank you @tm-bookshop 🙏

mickael-menu:
Ha, I made some changes and also fixed the lint but I can't push it to your fork. Could you enable "Edit from maintainers" in your PR? Thanks

tm-bookshop:
@mickael-menu diff applied, thank you. I think the "Edit from maintainers" option might be disabled because I'm on an Enterprise github account

alecdhansen:
@mickael-menu do you have a date estimate when we can get this released with 3.0.4?

mickael-menu:
This will be in the 3.1.0. I'd like to tackle a couple more issues before doing a release, but I don't have a fixed date yet.

PR #628, opened 2025-03-21
Modify the parsing of the guide node in EPUB 2.0 to return the same results as landmarks in EPUB 3.x.

erkasraim:
In my app, we use EPUB files created with the EPUB 2.0 format, which require the `guide` node. 
However, the library currently ignores the `guide` node. 

So, I modified it to return the same results as `landmarks` in the EPUB 3.x specification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file. Take a look
 
 ### Added
 
-* The EPUB 2 `<guide>` element is now parsed into the RWPM `landmarks` subcollection when no EPUB 3 `landmarks` navigation document is declared.
+* The EPUB 2 `<guide>` element is now parsed into the RWPM `landmarks` subcollection when no EPUB 3 `landmarks` navigation document is declared (contributed by [@erkasraim](https://github.com/readium/kotlin-toolkit/pull/628)).
 
 ### Changed
 


readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
```diff
@@ -34,6 +37,7 @@ internal data class PackageDocument(
                 ?: return null
             val spineElement = document.getFirst("spine", Namespaces.OPF)
                 ?: return null
+            val guideElement = document.getFirst("guide", Namespaces.OPF)
```
qnga, line 40: For the sake of consistency, you could get rid of the `null` case here.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/ManifestAdapter.kt
```diff
@@ -59,14 +59,20 @@ internal class ManifestAdapter(
             }
             .mapValues { listOf(PublicationCollection(links = it.value)) }
 
+        val guide = if (epubVersion < 3.0) {
```
mickael-menu, line 62: According to the spec, we should parse the `guide` element even in EPUB 3.0, but only if `landmarks` are missing.

> EPUB 3 Reading Systems must ignore the `guide` element when provided in EPUB 3 Publications whose [EPUB Navigation Document](https://idpf.org/epub/30/spec/epub30-publications.html#gloss-content-document-epub-nav) includes the `landmarks` feature.
> https://idpf.org/epub/30/spec/epub30-publications.html#sec-guide-elem

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
```diff
@@ -106,6 +111,26 @@ internal data class Spine(
...
+        // Epub 3.0+ does not support the guide element
+        // https://idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#TOC2.6
+        fun parse(element: ElementNode?, epubVersion: Double): List<Link> {
+            if (element == null || epubVersion >= 3.0) return emptyList()
+
+            return element.get("reference", Namespaces.OPF).mapNotNull { node ->
+                Link(
+                    href = node.getAttr("href")?.let { Url.fromEpubHref(it) } ?: return@mapNotNull null,
+                    title = node.getAttr("title"),
+                    properties = Properties().add(mapOf("type" to (node.getAttr("type") ?: "")))
```
qnga, line 127: Do you need this? We currently don't parse the type from EPUB 3 landmarks so we don't have a proper modeling approach for now.

mickael-menu (commented):
Thank you for your contribution!

I think that makes a lot of sense, [as `landmarks` is the replacement of the deprecated `guide`](https://idpf.org/epub/30/spec/epub30-publications.html#sec-guide-elem).

To make it more useful for integrators, I suggest some changes:

- `epub:type` is currently not parsed in EPUB 3 `landmarks`, so it's not very useful. I suggest [parsing it](https://github.com/readium/kotlin-toolkit/blob/dc25e7bf802bbb8bb32d796d8df8a3a5444ebe04/readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/NavigationDocumentParser.kt#L63) (+ [test cases](https://github.com/readium/kotlin-toolkit/blob/a226128e15d2a3052d6b8d7f5db8252bb630bd76/readium/streamer/src/test/java/org/readium/r2/streamer/parser/epub/NavigationDocumentParserTest.kt#L136-L141)) using the link property key `http://www.idpf.org/2007/ops#type` as it's an EPUB extension and not RWPM-native.
- We should map the EPUB 2 types to the EPUB 3 ones, so that integrators can use the `landmarks` without checking for the EPUB types. Here's some mapping suggestions. I highlighted the ones that are different.

  | [EPUB 2](https://idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section2.6) | [EPUB 3](https://www.w3.org/TR/epub-ssv-11/) |
  |--------|--------|
  | `title-page` | **`titlepage`** |
  | `text` | **`bodymatter`** |
  | `acknowledgements` | **`acknowledgments`** (note diff orthography) |
  | `notes` | **`endnotes`** (or `footnotes`?) |
  |`cover`|`cover`|
  | `toc ` | `toc` |
  | `index` | `index` |
  | `glossary` | `glossary` |
  | `bibliography` | `bibliography` |
  | `colophon` | `colophon` |
  | `copyright-page` | `copyright-page` |
  | `dedication` | `dedication` |
  | `epigraph` | `epigraph` |
  | `foreword` | `foreword` |
  | `loi` | `loi` |
  | `lot` | `lot` |
  | `preface` | `preface` |

mickael-menu:
I opened a related discussion: https://github.com/readium/webpub-manifest/discussions/111

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
erkasraim: Since the guide node may be missing, I handled it as nullable.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
erkasraim: In the app I develop and maintain, there is a feature that skips the cover and table of contents, displaying the main content first. In this case, it is necessary to check whether the text type is set in the guide

I will revise it again based on @mickael-menu 's suggestion, following the EPUB 3.x standard.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/ManifestAdapter.kt
erkasraim: fix it. Thanks

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
```diff
@@ -106,6 +111,40 @@ internal data class Spine(
...
+
+            return element.get("reference", Namespaces.OPF).mapNotNull { node ->
+                val href = node.getAttr("href")
+                    ?.let { Url.fromEpubHref(it) }
+                    ?.let { filePath.resolve(it) }
+                    ?: return@mapNotNull null
+                Link(
+                    href = href,
+                    title = node.getAttr("title"),
+                    rels = setOf(mapToEPUB3Spec(node.getAttr("type") ?: "")),
```
qnga, line 131: Please don't add an empty rel here. It's hard to ensure that every caller filters the values.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
qnga: I meant with `?: emptyList()`, not `return null`.  Not that important though.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/NavigationDocumentParser.kt
```diff
@@ -76,15 +76,34 @@ internal object NavigationDocumentParser {
...
+        val typeAttr = first.getAttrNs("type", Namespaces.OPS) ?: ""
+        val rels = if (typeAttr.isNotEmpty()) {
+            val types = parseProperties(typeAttr).map {
+                resolveProperty(
+                    it,
+                    prefixMap,
+                    DEFAULT_VOCAB.TYPE
+                )
+            }
+            types.map { it.removePrefix(Vocabularies.TYPE) }.toSet()
```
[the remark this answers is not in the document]
qnga: Actually, I would use a URL for every value for now, as suggested in https://github.com/readium/webpub-manifest/issues/50.

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
erkasraim: Since guideElement is an ElementNode, it cannot use ?: emptyList().

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
erkasraim: I have modified the code. Thank you!

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/PackageDocument.kt
```diff
@@ -106,6 +111,40 @@ internal data class Spine(
...
+                )
+            }
+        }
+
+        private fun mapToEPUB3Spec(type: String): String {
+            return when (type) {
+                "title-page" -> "titlepage"
+                "text" -> "bodymatter"
+                "acknowledgements" -> "acknowledgments" // American English
+                "notes" -> "endnotes" // endnotes or footnotes. https://www.w3.org/TR/epub-ssv-11/#notes
```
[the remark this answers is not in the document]
erkasraim: I have modified the code. Thank you!

readium/streamer/src/main/java/org/readium/r2/streamer/parser/epub/NavigationDocumentParser.kt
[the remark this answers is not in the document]
erkasraim: I have modified the code. Thank you!

qnga (approved):
I think we're good now.

mickael-menu (approved):
Thank you for making the changes @erkasraim 🙏

PR #629, opened 2025-03-22
ci: use jitpack instead of github packages

shovel-kun:


shovel-kun:
Sorry, this was a mistake.

PR #634, opened 2025-03-25
Add support for TDM Reservation Protocol metadata

mickael-menu:
#### Shared

* Support for [W3C's Text & data mining Reservation Protocol](https://www.w3.org/community/reports/tdmrep/CG-FINAL-tdmrep-20240510/) in our metadata models.


qnga (approved):
Looks good

PR #640, opened 2025-04-13
Update Room and enable KSP2

stevenzeck:


mickael-menu (approved):
Looking good, thank you.

PR #642, opened 2025-04-15
fix: webpub EPUBs

### LINKED ISSUE #641: [Bug] WebPubs don't seem to work in v3
### Describe the bug

I was trying to upgrade our kotlin-toolkit version from v2.4.0 to v3.0.3 and ran into an issue with our books in WebPublication format.
We use this format, so that we can stream and/or download the book depending on size and user preference.

The book opens and HTML content is shown, but ReadiumCSS never seems to be injected. Log shows a bunch of these errors:
```
[INFO:CONSOLE(1)] "Uncaught ReferenceError: readium is not defined"
```
Page turning and scrolling behaves strangely and changes to presentation does not seem to apply in any way. I'm assuming something goes wrong with the ReadiumCSS injection in the webview.

Attached is a .webpub (renamed to .zip so I could upload), of your own example webpub from the readium/webpub-manifest repo. If you need one of our specific .webpubs which behaves similarly please contact me on your Slack.

See also screenshot of Moby Dick open, having tried to change the bg-color to yellow. It only applies behind the WebView.
Image

### How to reproduce?

1. Open any .webpub book
2. See that HTML/text is shown, but no styling is applied and javascript cannot be evaluated.

### Readium version

v3.0.3

### Android API version

Latest

### Additional context

[rwp_ref_moby.zip](https://github.com/user-attachments/files/19737342/rwp_ref_moby.zip)

ddfreiling:
Fixes #641

Unlike swift-toolkit, this one did not set the correct `positionsServiceFactory` for WebPubs which conform to EPUB (regular EPUBs packaged as WebPub).

See https://github.com/readium/swift-toolkit/blob/develop/Sources/Streamer/Parser/Readium/ReadiumWebPubParser.swift#L108-L109

There also seemed to be an issue with FXL being the default layout, if none set via metadata.presentation. This does not seem to be the case with other toolkits and broke CSS/JS injections for our WebPubs. @mickael-menu maybe there is a smarter way to fix this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ All notable changes to this project will be documented in this file. Take a look
 
 * [go-toolkit#92](https://github.com/readium/go-toolkit/issues/92) The accessibility feature `printPageNumbers` is deprecated in favor of `pageNavigation`.
 
+### Fixed
+
+* Fixed support of Readium Web Publication packages conforming to the EPUB profile (contributed by [@ddfreiling](https://github.com/readium/kotlin-toolkit/pull/642)).
+
 
 ## [3.0.3]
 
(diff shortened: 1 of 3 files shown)


mickael-menu (approved):
Looking good, thank you for fixing this and making the changes Daniel!

PR #644, opened 2025-04-18
W3C Accessibility Metadata Display Guide

mickael-menu:
#### Shared

* Implementation of the [W3C Accessibility Metadata Display Guide](https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/guidelines/) specification to facilitate displaying accessibility metadata to users. [See the dedicated user guide](https://github.com/readium/kotlin-toolkit/blob/a11y-display-guide/docs/guides/accessibility.md).


readium/shared/src/main/java/org/readium/r2/shared/publication/Accessibility.kt
```diff
@@ -316,6 +337,11 @@ public data class Accessibility(
     public data class Feature(public val value: String) : Parcelable {
 
         public companion object {
+            /**
+             * Indicates that the resource does not contain any accessibility features.
+             */
+            public val NONE: Feature = Feature("none")
```
qnga, line 343: I guess you're just mimicking what we've got with hazards, but this reminds me a bit of the C language. `None` is not a feature and there are other ways to express the absence of something in Kotlin. We could have:
- the empty set denoting the absence of any feature
- the null value denoting the absence of data about features

Though showing `None` in a feature list would be an acceptable UI, a better one would display it in a different style than true features IMHO. A good UI could do as follows:
- if `features == null` don't do anything
- else if features.isEmpty() explain that this publication doesn't contain anything special (for instance with a small text under the title `Features`)
- else display the list of features

mickael-menu: Although I agree with your general comment, in this specific case [`none` is a value for `accessibilityFeature`](https://www.w3.org/community/reports/a11y-discov-vocab/CG-FINAL-vocabulary-20241209/#feature-none). It's different from an empty set or no metadata declared. It can be used by the author to explicitly declare that no accessibility features are provided.

(in this PR I just moved it at the top of the enum cases to clarify the comment sections, it was already there)

qnga: What would an empty set explicitly provided (as opposed to `null`) convey if that's not "no accessibility features are provided"? I think the `none` value in the accessibility spec is just a way of encoding the underlying concept. We're not bounded to encode concepts in the same way in a different (and more expressive) language.

qnga: Anyway, that's only a move, indeed. We can possibly revisit this later.

mickael-menu: I'd rather stay consistent with the other Accessibility models (`Hazards`, which has `none` but also `unknown`) and specifications ([RWPM](https://github.com/readium/webpub-manifest/tree/master/contexts/default#features-and-hazards), where your argument can also apply). This API is a pretty direct mapping.

The accessibility display guide introduced in this PR is the higher-level API which should be used to build a UI around the accessibility metadata. Did you see anything that could be improved in the display guide API?

qnga: Consistency is a good point. Actually, I would probably have written the hazards more or less like you did in the accessibility display guide. That's a big question: stay close to simple serialization languages or not? In simple data classes such as `Manifest`, that could make some sense. On the other hand, that's neither idiomatic nor convenient for API consumers. We don't have a higher-level API for everything.

The accessibility display guide looks pretty good. I'll have a more in-depth look later.

qnga (commented):
Looks great. What about putting the display guide and related stuff into an `accessibility` package at the module root though? Of course this is a e-reading toollkit so ultimately everything is related to publications. We could keep the `publication` package for the components of a `publication` themselves.

mickael-menu:
@qnga I agree, I moved the guide into its own `accessibility` package.

PR #664, opened 2025-05-15
fix: corrected typo 'spread-node' to 'spread-none'

BienBoy:
The incorrect rendition property 'spread-node' has been fixed to the correct 'spread-none' to ensure proper parsing of EPUB rendition metadata.


qnga (approved):
Oh dear! Thanks for that.

PR #678, opened 2025-05-31
Update dependencies and Android SDK

stevenzeck:
I'll update the README before this is ready to merge. I would say wait until Android 16 starts pushing out to devices.

**Note**
Updating jsoup causes the tests in [HtmlResourceContentIteratorTest](https://github.com/readium/kotlin-toolkit/blob/38b1a810235409967b9c9bacea1516aabc344699/readium/shared/src/test/java/org/readium/r2/shared/publication/services/content/iterators/HtmlResourceContentIteratorTest.kt) to fail, as the CSS selectors in the [elements](https://github.com/readium/kotlin-toolkit/blob/38b1a810235409967b9c9bacea1516aabc344699/readium/shared/src/test/java/org/readium/r2/shared/publication/services/content/iterators/HtmlResourceContentIteratorTest.kt#L53-L159) list no longer match the ones parsed by jsoup in the [HTML](https://github.com/readium/kotlin-toolkit/blob/38b1a810235409967b9c9bacea1516aabc344699/readium/shared/src/test/java/org/readium/r2/shared/publication/services/content/iterators/HtmlResourceContentIteratorTest.kt#L32-L51). I believe this is due to https://github.com/jhy/jsoup/pull/2283. I've reverted it for now because I don't know if there are other implications elsewhere.


mickael-menu (approved):
Thanks Steven 🙏

PR #682, opened 2025-06-12
feat: support TTSNavigator and text search for WebPubs with HTML content

ddfreiling:
Add missing content and search services to WebPub publications containing HTML.

Ensuring that TTSNavigator and text search works for these types of publication.


mickael-menu (approved):
Thank you @ddfreiling, this is helpful!

PR #699, opened 2025-08-01
Refactor R2RTLViewPager using ViewPager2 for better RTL and scroll behavior

erkasraim:
I’ve migrated R2RTLViewPager to R2RTLViewPager2.
The key changes are as follows:

1. Applying RTL is now much simpler with ViewPager2.

2. With ViewPager2's support for both horizontal and vertical orientation,
WebViews can now be laid out vertically in scroll mode, which provides a better user experience.

3. In R2RTLViewPager2’s onInterceptTouchEvent(),
touch events are controlled based on the WebView's scroll state,
allowing for smoother swipe handling when switching between pages containing WebViews.


erkasraim:
I’ll be closing this PR because of issues with the commit history and will open a new one.

Issue #709, opened 2025-09-03
Future of the PDFium adapter in Readium (16 KB page sizes issue)


mickael-menu:
As you probably know:

> Starting [November 1st, 2025](https://android-developers.googleblog.com/2025/05/prepare-play-apps-for-devices-with-16kb-page-size.html), all new apps and app updates that use native C/C++ code targeting Android 15+ devices submitted to Google Play must support 16 KB page sizes. This is a crucial step towards ensuring your app delivers the best possible performance on the latest Android hardware.
> https://android-developers.googleblog.com/2025/07/transition-to-16-kb-page-sizes-android-apps-games-android-studio.html

LCP is ready for 16 KB page sizes but we still use an old unmaintained dependency which is not compatible in the Readium toolkit: PDFium.

Unfortunately, core maintainers are not able to allocate time to maintain the PDF adapter for PDFium to handle 16 KB page sizes. If you are interested in preserving this support in the Readium toolkit, we encourage you to submit a PR to implement it.

Apparently, a few forks exist which support 16 KB page sizes:

- https://github.com/AhmerAfzal1/AhmerPdfium
- https://github.com/v-mas/PdfiumAndroid/
- https://github.com/DImuthuUpe/AndroidPdfViewer/issues/1191#issuecomment-2893824822

The tricky part is that, [due to a conflict with liblcp](https://github.com/readium/kotlin-toolkit/issues/29), the Readium toolkit is stuck with the PdfiumAndroid 1.8.2 codebase, which may not be updated for 16 KB in these forks.

**If this is important to you and you would like to contribute to this update, please come forward.** Otherwise, we will remove support for PDFium in early October, to be prepared for the November deadline.

qnga:
Fixed by #710

mickael-menu:
[3.1.2 is released](https://github.com/readium/kotlin-toolkit/releases/tag/3.1.2) with support for 16 KB page sizes.

PR #710, opened 2025-09-04
Fix 16K support for PDFium

### LINKED ISSUE #638: [Bug] 16 KB compatibility
### Describe the bug

When running against a 16 KB Page Size virtual device:

> This app isn't 16 KB compatible. ELF alignment check failed. This app will be run using page size compatible mode. For best compatibility, please recompile the application with 16 KB support. For more information, see https://developer.android.com/guide/practices/page-sizes.

### How to reproduce?

1. Create a virtual device using an 16 KB image.
2. Run the app against that device.

Result: That above message displays.

### Readium version

develop

### Android API version

35

### Additional context

_No response_

gustavaa:
Fixes #638, #709 

This PR fix the 16K support issue with readium using the forks of marain87

[Pdfium Android fork](https://github.com/marain87/PdfiumAndroid)
[AndroidPdfViewer fork](https://github.com/marain87/AndroidPdfViewer)

These are compatible with each other and are both published on jitpack. 

I have verified that PDFs are working in the test app on a 16k emulator without any warning messages about 16k compatibility. I have also built a version of the readium-toolkit library locally and tested it in our production apps, both with and without LCP and have not run into any issues.

As `:readium:adapters:pdfium:readium-adapter-pdfium-common` only acted as an umbrella for the locally included Pdfium lib, I removed it and replaced the references with a dependency directly on the new remote lib instead.


qnga (approved):
Looks to work great, thanks again for tackling this!

PR #716, opened 2025-10-07
fix(tts): add missing supported timeline command to TtsSessionAdapter

ddfreiling:
without this command the session-adapter didn't properly report the current timeline to a media3.MediaSession

### DIFF
--- a/readium/navigators/media/tts/src/main/java/org/readium/navigator/media/tts/session/TtsSessionAdapter.kt
+++ b/readium/navigators/media/tts/src/main/java/org/readium/navigator/media/tts/session/TtsSessionAdapter.kt
@@ -157,7 +157,8 @@ internal class TtsSessionAdapter<E : TtsEngine.Error>(
                 COMMAND_SET_SPEED_AND_PITCH,
                 COMMAND_GET_CURRENT_MEDIA_ITEM,
                 COMMAND_GET_METADATA,
-                COMMAND_GET_TEXT
+                COMMAND_GET_TEXT,
+                COMMAND_GET_TIMELINE
             ).build()
 
     override fun getApplicationLooper(): Looper {
--- a/readium/navigators/media/tts/src/main/java/org/readium/navigator/media/tts/session/TtsSessionAdapter.kt
+++ b/readium/navigators/media/tts/src/main/java/org/readium/navigator/media/tts/session/TtsSessionAdapter.kt
@@ -458,11 +458,16 @@ internal class TtsSessionAdapter<E : TtsEngine.Error>(
 
     override fun getCurrentTimeline(): Timeline {
         // MediaNotificationManager requires a non-empty timeline to start foreground playing.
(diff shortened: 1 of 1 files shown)


qnga:
What is timeline used for?

ddfreiling:
If using the [ForwardingSimpleBasePlayer](https://developer.android.com/reference/androidx/media3/common/ForwardingSimpleBasePlayer) to modify MediaSession, it has a `getState` function to build the player state, which depends on the adapter player allowing the `COMMAND_GET_TIMELINE` to be called, which currently `TtsSessionAdapter` does not.
Screenshot 2025-10-07 at 13 57 15

qnga:
I understand that. But what does this getState do better with the timeline? I'm not entirely sure my timeline implementation is good or useful in the adapter, that's why I ask.

ddfreiling:
Without a timeline the SimpleBasePlayer dies on an assert that timeline must not be empty if the player is not in IDLE state. For us this crashes the entire MediaSession when we call `openSession` on it with a TtsNavigator.

Exception:
```
SimpleBasePlayer: Empty playlist only allowed in STATE_IDLE or STATE_ENDED
```

As long as SimpleBasePlayer sees a non-empty timeline, everything seems to work.

I guess this only happens, since we wrap the `navigator.asMedia3Player()` in a `ForwardingSimpleBasePlayer` to better control certain media-button events and notification button layout. Maybe there's another way to do this?

qnga:
Ok, I see. If you want a specific notification button layout, you might expect a better timeline than my shitty one. Are you expecting resource-based navigation in the notification? I might have not exposed it because of the notification actually, I need to check.

qnga:
As far as I understand and remember the missing timeline command is more of a feature than a bug. To prevent your crash you should be able to provide a timeline by yourself by overriding `getState` or something of that ilk. If for some some reason that's not enough, please report your use case in detail.

ddfreiling:
@qnga turns out I have to completely duplicate/re-implement the entire `getState` method of ForwardingSimpleBasePlayer. Because if I call super.state and just expand on that, I still get the assert exception:
```
Empty playlist only allowed in STATE_IDLE or STATE_ENDED
```
So basically I have to copy their implementation but leave out this non-empty playlist assert and the check on `AvailableCommands.COMMAND_GET_TIMELINE`.

Maybe the TtsSessionAdapter should just report a single item on the playlist when active?

qnga:
A single-item timeline sounds reasonable. As mentioned above, my main concern is about the notification. Any solution not making the backward and forward buttons appear in the default notification would be fine with me as this was the intended behavior.

ddfreiling:
What do you think about this solution? In the test-app I'm seeing a simple notification with just a pause/play button now, which also supports wrapping the player in a ForwardingSimpleBasePlayer.

qnga:
That looks fine!

qnga (approved):

PR #717, opened 2025-10-08
Update dependencies and JVM target

qnga:
Update dependencies including AGP and bump JVM target  to 11 (1.8 is deprecated).

### DIFF
--- a/readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2ViewPager.kt
+++ b/readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2ViewPager.kt
@@ -14,9 +14,12 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.MotionEvent
 import org.readium.r2.navigator.BuildConfig.DEBUG
+import org.readium.r2.shared.InternalReadiumApi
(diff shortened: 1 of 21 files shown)


readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2ViewPager.kt
```diff
@@ -14,9 +14,12 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.MotionEvent
 import org.readium.r2.navigator.BuildConfig.DEBUG
+import org.readium.r2.shared.InternalReadiumApi
 import timber.log.Timber
 
-internal class R2ViewPager : R2RTLViewPager {
+// See https://youtrack.jetbrains.com/issue/KTLC-271 for visibility issue.
+@InternalReadiumApi
```
qnga, line 21: Kotlin/Java interop isn't as good as we thought. This is the easiest workaround.

stevenzeck:
At some point we'll have to address the other annotated parameters, per https://youtrack.jetbrains.com/issue/KT-73255. Otherwise looks good to me.

mickael-menu (approved):

PR #719, opened 2025-10-17
Prevent multiple OutlineFragment instances on repeated menu clicks.

janiceferns02:
This PR fixes the issue where multiple OutlineFragment instances are added when the menu item is clicked repeatedly in quick succession in the ReaderActivity.

Screenshot 2025-10-17 at 3 38 56 PM

Fix: Before adding OutlineFragment, we now check if it already exists in the FragmentManager using findFragmentByTag.

Tested by repeatedly tapping the ≡ menu icon multiple times very quickly and confirmed that only a single instance of OutlineFragment is ever created.


mickael-menu:
Thank you @janiceferns02. Could you address the formatting issue? You can run `make format` in the project to do so.

janiceferns02:
@mickael-menu I have resolved the formatting issue. Am I still missing something?

mickael-menu (approved):
Thank you for making the changes @janiceferns02

PR #720, opened 2025-10-24
Allow custom LCP device identifiers

dewantawsif:
Also make sure that we set generated device id only once after `DeviceService` has been initialized


mickael-menu (approved):
Thank you @dewantawsif

PR #727, opened 2025-11-21
Fix File.toUrl and DirectoryContainer

qnga:
`DirectoryContainer` didn't work on real devices though the unit tests worked fine because of a discrepancy between the Android and regular JRE implementations of `URI` (see #725).
Now, the caller must specify if a file is a directory or a regular file when converting a `File` to `Url` so that `relativize` works fine.


mickael-menu (approved):
Looking good, thank you for making the changes.

PR #738, opened 2026-01-18
Update Gradle, AGP, and other dependencies

stevenzeck:
- Updates Gradle from 8.14.1 to 9.1.0
- Updates the Android Gradle Plugin from 8.11.1 to 9.0.0. In doing this, I migrated it to built-in Kotlin. That means that the`id("org.jetbrains.kotlin.android")` plugin is no longer needed in the build.gradle.kts files. The `compilerOptions.jvmTarget` block is also no longer needed because its value defaults to android.compileOptions.targetCompatibility.
- The AGP update also introduced several new flags available in gradle.properties.
- Updated dependencies of other libraries (note that KSP is no longer tied to the Kotlin version anymore).
- Fixed deprecations and API changes due to the dependencies updates.
- Modified the root build.gradle.kts to register the clean task everywhere it doesn't have one. Even before this update, using Android Studio's `clean` resulted in errors like this: "Cannot locate tasks that match ':demos:clean' as task 'clean' not found in project ':demos'."

### DIFF
--- a/.idea/kotlinc.xml
+++ b/.idea/kotlinc.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="KotlinJpsPluginSettings">
-    <option name="version" value="2.0.21" />
+    <option name="version" value="2.2.0" />
   </component>
 </project>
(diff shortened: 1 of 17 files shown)


qnga:
Please keep the toolkit compatible as a subbuild with Kotlin multiplatform as it is the way I develop it at the moment.
https://kotlinlang.org/docs/multiplatform/multiplatform-compatibility-guide.html#version-compa

PR #739, opened 2026-01-18
Update Kotlin, Gradle, and other dependencies

stevenzeck:
- Updates Gradle from 8.14.1 to 9.0.0
- Updates the Android Gradle Plugin from 8.11.1 to 8.13.2.
- Updated dependencies of other libraries (note that KSP is no longer tied to the Kotlin version anymore).
- Fixed deprecations and API changes due to the dependencies updates.
- Modified the root build.gradle.kts to register the clean task everywhere it doesn't have one. Even before this update, using Android Studio's `clean` resulted in errors like this: "Cannot locate tasks that match ':demos:clean' as task 'clean' not found in project ':demos'."


gradle/libs.versions.toml
```diff
@@ -1,35 +1,35 @@
 [versions]
 
-kotlin = "2.2.20"
-agp = "8.11.1"
+kotlin = "2.3.0"
+agp = "8.13.2"
```
qnga, line 4: 8.13.0 max for multiplatform compat please.

gradle/libs.versions.toml
```diff
@@ -1,35 +1,35 @@
...
 androidx-annotation = "1.9.1"
 androidx-appcompat = "1.7.1"
 androidx-browser = "1.9.0"
 androidx-cardview = "1.0.0"
-androidx-compose-animation = "1.9.2"
-androidx-compose-foundation = "1.9.2"
-androidx-compose-material = "1.9.2"
+androidx-compose-animation = "1.10.1"
+androidx-compose-foundation = "1.10.1"
+androidx-compose-material = "1.10.1"
```
qnga: 1.10.0 max for multiplatform compat please.

gradle/libs.versions.toml
```diff
@@ -46,13 +46,10 @@ junit = "4.13.2"
 kotlinx-coroutines = "1.10.2"
 kotlinx-coroutines-test = "1.10.2"
 kotlinx-datetime = "0.7.1"
-kotlinx-serialization-json = "1.8.1"
+kotlinx-serialization-json = "1.9.0"
```
qnga: This should be 1.10.0 so that it doesn't conflict with kotlinx-coroutines.

qnga (approved):
Looks good

PR #740, opened 2026-01-20
Update Dokka, fix documentation generation

stevenzeck:
This PR updates the Dokka plugin and makes changes to the documentation generation. Right now, at least the way it's configured even though it hasn't been run, the Dokka task generates KDocs in markdown format. From there, the workflow uses mkdocs to combine those KDocs with the `docs` directory and generates the HTML.

This PR changes that. Now, Dokka will generate HTML for the KDocs. When mkdocs runs, it puts a link to the kdocs part of the site in the navigation. We lose the ability to switch versions in the header for Dokka's KDocs (the handwritten docs are not affected by this), but I think that's a small price to pay for having cleaner, better looking KDocs.

I also fixed some of the links in mkdocs.yml to point to the new file locations, but otherwise didn't make any changes to the content.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,8 +4,6 @@
  * available in the top-level LICENSE file of the project.
  */
 
-import org.jetbrains.dokka.gradle.DokkaTaskPartial
-
 plugins {
     alias(libs.plugins.dokka)
     alias(libs.plugins.ktlint)
@@ -13,7 +11,9 @@ plugins {
 }
 
 subprojects {
-    if (name != "test-app") {
+    val shouldDocument = name != "test-app" && !path.startsWith(":demos")
+
+    if (shouldDocument) {
         apply(plugin = "org.jetbrains.dokka")
     }
     apply(plugin = "org.jlleitschuh.gradle.ktlint")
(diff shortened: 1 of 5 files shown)


stevenzeck:
@mickael-menu I remember there were issues with running the docs workflow. I don't remember what they were exactly, but I think it was more of a GitHub configuration thing with the repository than the job itself.

build.gradle.kts
```diff
@@ -24,19 +24,31 @@ subprojects {
...
+    )
 }
 
-tasks.withType<DokkaTaskPartial>().configureEach {
-    dokkaSourceSets {
-        configureEach {
-            reportUndocumented.set(false)
-            skipEmptyPackages.set(false)
-            skipDeprecated.set(true)
+subprojects {
```
mickael-menu: There is already a `subprojects` section above, and the `shouldDocument` part is duplicated.

mkdocs.yml
```diff
@@ -10,7 +10,7 @@ repo_name: kotlin-toolkit
 repo_url: https://github.com/readium/kotlin-toolkit
 
 # Copyright (shown at the footer)
-copyright: 'Copyright &copy; 2023 Readium Foundation'
+copyright: 'Copyright &copy; 2026 Readium Foundation'
```
mickael-menu, line 13: Apparently we could use the macros plugin to keep this up to date automatically.

It probably requires adding the `pip install` command in the workflow: https://mkdocs-macros-plugin.readthedocs.io/en/latest/

```yaml
plugins:
  - macros

# Copyright (shown at the footer)
copyright: 'Copyright &copy; {{ now().year }} Readium Foundation'
```

stevenzeck: We can, but I don't think that docs are run often enough to warrant using it just for that.

mickael-menu: I'm more concerned that we will likely forget to update the date. I'll merge this and see if we can get the GitHub workflow running before adjusting the docs and that kind of stuff.

mickael-menu (approved):

PR #741, opened 2026-01-20
Remove title inference based on folder names for raw archives

mickael-menu:
#### Streamer

*  Removed title inference based on folder names within image and audio archives. Use the archive's filename instead.

---

This title inference was not a well-known or common convention in the comic book archive community.


qnga:
Have you conducted a large-scale investigation? I'm not sure about comics but the pattern is quite largely spread for audiobooks.
Anyway, what do we have instead? Are you sure the archive filename is better than that? I can't imagine a root directory name that would be worse.

mickael-menu:
I was the one implementing this logic after observing it in our CBZ test case (Futuristic Tales), but:

- I cannot find a reference for this convention anywhere, and other Comic Book apps are not doing this.
- It is problematic in practice when CBZ files sold on the market contain garbage folder names, such as `working_files`, which have been observed in the wild.

> the pattern is quite largely spread for audiobooks

Do you have sources for this claim? And/or audiobook apps expecting it?

mickael-menu:
Looking for audiobooks samples online, I found only audiobooks with no folders, and one with a crappy folder `Claude_Prosper_Jolyot_de_Crébillon-Lettres_de_la_marquise_de_M_au_comte_de_R` (https://www.litteratureaudio.com/livre-audio-gratuit-mp3/claude-prosper-jolyot-de-crebillon-lettres-de-la-marquise-de-m-au-comte-de-r.html).

smoores-dev:
Just adding some more data points from what we've seen/support with Storyteller:

- Libro.fm (major DRM-free audiobook provider) provides audiobooks as either a single flat M4B  (MPEG4) file or 1 or more zip archives of MP3 files. The zip archives have no root folder, just a flat list of MP3 files. Conceivably a user might unzip these and then re-zip them into a new archive with multiple top level folders, one for each originial archive (in which case the names would not be helpful).
- There are popular DRM-stripping tools that allow users to remove DRM from their own files from Audible, etc. Those usually output flat MP3/MP4 files.

We don't see a lot of publishers/sources providing archives generally, other than to collect a flat list of MP3 files. Storyteller supports users uploading or importing 1 or more flat audio files, or arbitrarily nested folders and ZIP archives, but we only use the top level folder or archive name as the fallback title.

Also, we rarely actually fall back to that title, since almost all sources at the very least add album (usually book title) and/or title (book title or chapter title) metadata to the audio file itself, which is where we look for it first

mickael-menu:
Thank you @smoores-dev, that's useful feedback.

Note that implementers can still generate a fallback title from the root folder using the Readium `Container` API if necessary. I argue that including this in `publication.metadata.title` is misleading. Since it is not a fallback title extracted from real metadata, we cannot vet it.

qnga:
> I was the one implementing this logic after observing it in our CBZ test case (Futuristic Tales), but:

In my memories, it was me after observing it in audiobooks.

>  I cannot find a reference for this convention anywhere

That's what a convention or _de facto_ standard is.
 
> Looking for audiobooks samples online, I found only audiobooks with no folders, and one with a crappy folder 

Same result for me.

>  It is problematic in practice when CBZ files sold on the market contain garbage folder names, such as `working_files`, which have been observed in the wild.

Ok

>  I argue that including this in publication.metadata.title is misleading. Since it is not a fallback title extracted from real metadata, we cannot vet it.

It was added at a time when we used fallbacks. I agree that title fallbacks are a bad idea anyway for that reason. Looking at your PR description, I first thought you were using the archive filename as a fallback. Looking at the code, it looks like you don't.

qnga (approved):

PR #747, opened 2026-01-30
Improve new Web navigators

qnga:
- Fix a bunch of bugs
- Upgrade Readium CSS
- Improve progression computation and restoration


readium/navigators/common/src/main/java/org/readium/navigator/common/LocationElements.kt
```diff
@@ -60,3 +79,39 @@ public data class TextQuote(
...
+@ExperimentalReadiumApi
+public fun TextQuote.toTextAnchor(end: Boolean = false): TextAnchor =
+    when (end) {
+        false -> TextAnchor(
+            textBefore = prefix,
+            textAfter = text + suffix
+        )
+        true -> TextAnchor(
+            textBefore = prefix + text,
+            textAfter = text
```
mickael-menu: When is it helpful to choose `end: true` instead of `false`?

```suggestion
            textAfter = suffix
```

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebPublication.kt
```diff
@@ -36,10 +43,35 @@ internal class ReflowableWebPublication(
...
+
+    fun positionForProgression(index: Int, progression: Progression): Position {
+        val itemPositionNumber = readingOrder.positionNumbers[index]
+        val localPosition = floor(progression.value * itemPositionNumber).toInt()
+        return Position(startPositions[index] + localPosition)!!
+    }
+
+    fun totalProgressionForPosition(position: Position): Progression {
+        return Progression(position.value / totalPositionCount.toDouble())!!
+    }
```
mickael-menu, line 78: I'm a bit uncomfortable with the force unwrapping here. It doesn't look like we can guarantee they will never crash.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebRenditionState.kt
```diff
@@ -277,42 +341,73 @@ internal class ReflowableLayoutDelegate(
...
+                            resourceStates.zip(destLocationByResource)
+                                .forEach { (state, location) ->
+                                    state.go(
+                                        location = location,
+                                        continuation = continuation.takeIf { state === resourceStates[destIndex] }
+                                    )
+                                }
+                        }
+                    } catch (_: Exception) {
+                        resourceStates.zip(destLocationByResource)
```
mickael-menu: I would at least log any exception here, as they are not really handled.

readium/navigators/web/internals/src/main/kotlin/org/readium/navigator/web/internals/webview/WebViewScrollController.kt
```diff
@@ -167,29 +192,61 @@ private fun RelaxedWebView.scrollToProgression(
...
+        LayoutDirection.Ltr -> scrollX / (maxScrollX + width).toDouble()
+        LayoutDirection.Rtl -> 1 - scrollX / (maxScrollX + width).toDouble()
+    }
+}
+
+private fun RelaxedWebView.endProgression(
+    orientation: Orientation,
+    direction: LayoutDirection,
+): Double {
+    Timber.d("endProgression $scrollX $width $maxScrollX")
```
mickael-menu:
```suggestion
```

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebRenditionState.kt
```diff
@@ -277,42 +341,73 @@ internal class ReflowableLayoutDelegate(
...
-            resourceStates[resourceIndex].scrollController.value?.moveToProgression(it)
+        coroutineScope {
+            navigatorMutex.mutateWith(
+                receiver = this,
+                priority = MutatePriority.UserInput
+            ) {
+                withContext(Dispatchers.Main) {
+                    val destIndex = publication.readingOrder.indexOfHref(location.href) ?: return@withContext
+                    val destLocationByResource = location.toResourceLocations(destIndex, publication.readingOrder)
+                    Timber.d("destByResource $destLocationByResource")
```
mickael-menu:
```suggestion
```

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/resource/ReflowableResourceState.kt
```diff
@@ -4,24 +4,134 @@
...
+            direction = direction
+        )?.let { Progression(it) }
+            ?: return false
+
+        val endProgression = scrollController.endProgression(
+            orientation = orientation,
+            direction = direction
+        )?.let { Progression(it) } ?: return false
+
+        Timber.d("updateProgression $startProgression $endProgression")
```
mickael-menu, line 103:
```suggestion
```

readium/navigator/src/main/assets/_scripts/src/gestures.js
```diff
@@ -13,11 +13,19 @@ window.addEventListener("DOMContentLoaded", function () {
...
+      window.getSelection().isCollapsed
+    }`
+  );
+
   if (!window.getSelection().isCollapsed) {
     // There's an on-going selection, the tap will dismiss it so we don't forward it.
     return;
   }
 
+  console.log("Reporting tap");
```
mickael-menu:
```suggestion
```

readium/navigator/src/main/assets/_scripts/src/gestures.js
```diff
@@ -13,11 +13,19 @@ window.addEventListener("DOMContentLoaded", function () {
 });
 
 function onClick(event) {
+  console.log(
+    `selectionType ${window.getSelection().type} isCollapsed ${
+      window.getSelection().isCollapsed
+    }`
+  );
```
mickael-menu:
```suggestion
```

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/resource/ReflowableResource.kt
```diff
@@ -167,58 +184,178 @@ internal fun ReflowableResource(
...
+                                        orientation = orientation
+                                    )
+                                    offset?.let { offset ->
+                                        scrollController.moveToOffset(
+                                            offset = with(density) { offset.dp.roundToPx() },
+                                            snap = !scroll,
+                                            orientation = orientation,
+                                        )
+                                    }
+                                }
```
mickael-menu: These three branches are basically doing the same thing. It would be worth refactoring.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/css/ReadiumCssProperties.kt
```diff
@@ -562,7 +578,4 @@ private fun String.toCss(): String =
  * Converts a [Double] to a string literal with the given [unit].
  */
 private fun Double.toCss(unit: String): String =
-    NumberFormat.getNumberInstance(Locale.ROOT).run {
-        maximumFractionDigits = 2
-        format(this@toCss)
-    } + unit
+    String.format(Locale.ROOT, "%e%s", this, unit)
```
mickael-menu, line 581: I'm not sure the scientific notation is supported in CSS, and it's probably not necessary anyways.

Added `.2` to keep the previous behavior, but maybe that's not useful.

```suggestion
    String.format(Locale.ROOT, "%.2f%s", this, unit)
```

readium/navigators/common/build.gradle.kts
```diff
@@ -22,11 +22,11 @@ dependencies {
     api(project(":readium:readium-shared"))
     api(project(":readium:readium-navigator"))
 
+    api(libs.androidx.compose.foundation)
+    api(libs.kotlinx.coroutines.android)
     api(libs.kotlinx.collections.immutable)
+    api(libs.kotlinx.serialization.json)
```
mickael-menu: It's a bit of a shame that these become part of our own public API surface, but I guess there's no way around it. Have you seen how other Compose-based projects manage this?

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/layout/LayoutConstants.kt
```diff
@@ -0,0 +1,13 @@
+package org.readium.navigator.web.reflowable.layout
```
mickael-menu, line 7: Some of the new files are lacking the copyright headers.

We should probably look into a linting tool that would automate this...

mickael-menu (commented):
I didn't review everything, but we will review again the public APIs together when releasing the beta.

A lot of doc comments are also missing, I guess we can add them when the API is more stable. But if you could add some to help understand the major components that would be helpful 🙏

qnga:
It is still draft because I didn't solve all the issues due to the upgrade to Compose 1.10, but thanks for the review.

readium/navigators/common/src/main/java/org/readium/navigator/common/LocationElements.kt
qnga: I don't think I've got a use case for it at the moment. But conceptually, you can't  go from a `TextQuote` to a `TextAnchor` without deciding if you want to point at the beginning or the end of the quoted text.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebPublication.kt
qnga: A position greater than the number of positions? It's a basic precondition, I think we can afford a crash here. Especially as it's internal at the moment.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebRenditionState.kt
qnga: Mmm, not sure as the main target is `CancellationException`. I will document this.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/css/ReadiumCssProperties.kt
qnga: It is supported in CSS. I made this change because the previous implementation inherited from the legacy navigator has a bug: it adds thousand separators though they are not supported in CSS.

readium/navigators/common/build.gradle.kts
qnga: There is a way around, but I thought that was the cleanest option. If we do not add this, users have to explicitly add the dependencies to their project to be able to use our APIs. I couldn't see the point of forcing them to do that...

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebPublication.kt
qnga: To be clear, this doesn't depend on any external input.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebRenditionState.kt
qnga: Finally I rethrow the exceptions. I catch them only to get a chance to do some clean up.

readium/navigators/common/build.gradle.kts
qnga: I rationalized this a bit more. I've kept only Compose Foundation and Immutable Collections as this is true: they are used in public APIs. `api` dependencies are meant exactly for this I think.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebPublication.kt
stevenzeck: Can't the nullability be propagated to the callers? Though there are several force unwrappings in `ReflowableWebRenditionState` too.

qnga: The caller will have the same issue. We have to assume somewhere that positions are correct (and that's a correct assumption). It's simpler to do this in the narrowest scope.

readium/navigators/common/build.gradle.kts
mickael-menu: I checked how other popular libraries handle this. Both [`coil`](https://github.com/coil-kt/coil/blob/60bf88aad7d3d47a6368b242ced2e8b2dde8d565/coil-compose-core/build.gradle.kts#L22) and [`media3`](https://github.com/androidx/media/blob/630c1af455a16b8eb1ab36f49d82d9db799d7f5e/libraries/ui_compose/build.gradle#L55) use `api()` for their Compose dependencies. The convention is to isolate Compose in a module with a `-compose` suffix, but since our new navigator modules are inherently UI components, the current setup is fine IMO.

Keeping `api()` for Compose and immutable collections makes sense.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/css/ReadiumCssProperties.kt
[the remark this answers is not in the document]
mickael-menu: @qnga Any reason to use scientific notation instead of plain `%f`?

mickael-menu (approved):
@qnga Could you just review the comments left by Copilot before merging? Thank you

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/css/ReadiumCssProperties.kt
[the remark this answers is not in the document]
qnga: Yes, it's universally defined so you won't get any issue like thousand separators. I can't remember if I tested it or just hurried to get rid of all those inconsistencies,  but "%f" with root locale may have them.
Regarding Copilot's point, that would be true on a Web code targeting every browser. In our case it works on Android WebView so it's fine.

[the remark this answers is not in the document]
qnga: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/number#browser_compatibility

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebPublication.kt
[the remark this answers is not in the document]
qnga: Actually progression can't be 1.0. But okay on principle.
Regarding totalProgression, that sounds true. We should aim at the beginning of the position though the current implementation aims at the end.

readium/navigators/web/internals/src/main/kotlin/org/readium/navigator/web/internals/webview/WebViewScrollController.kt
```diff
@@ -167,29 +191,60 @@ private fun RelaxedWebView.scrollToProgression(
...
         }
         Orientation.Horizontal -> when (direction) {
             LayoutDirection.Ltr -> {
-                scrollTo(progression.roundToInt() * maxScrollX, scrollY)
+                scrollTo(ceil(progression * docWidth).roundToInt(), scrollY)
             }
             LayoutDirection.Rtl -> {
-                scrollTo((1 - progression).roundToInt() * maxScrollX, scrollY)
+                scrollTo((ceil(1 - progression) * docWidth).roundToInt(), scrollY)
             }
```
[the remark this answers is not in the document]
qnga: Correct

PR #750, opened 2026-02-03
Create listeners before init block to avoid initialization issues

RankoR:
`listeners` variable should be initialized before `init` block that may use them, because in some cases they are accessed right from `init` call - and are not initialized at this moment, which leads to crashes.


qnga (approved):
Sounds good, thanks!

PR #754, opened 2026-02-13
Remove Presentation hints

qnga:
#### Shared

The Presentation Hints properties are deprecated from the Readium Web Publication Manifest models. [See the official documentation](https://readium.org/webpub-manifest/profiles/epub.html#appendix-b---deprecated-properties).


readium/shared/src/main/java/org/readium/r2/shared/publication/epub/EpubLayout.kt
```diff
@@ -20,6 +21,7 @@ import org.readium.r2.shared.util.MapCompanion
  * Hints how the layout of the resource should be presented.
  * https://readium.org/webpub-manifest/schema/extensions/epub/metadata.schema.json
  */
+@Deprecated("This was removed from RWPM. You can still use the EPUB extensibility to access the original values.")
```
mickael-menu: Adding a `ReplaceWith` pointing to the new `Layout` would help migrations of apps doing stuff like `preferenceEditor.layout == EpubLayout.REFLOWABLE`.

readium/shared/src/main/java/org/readium/r2/shared/publication/presentation/Metadata.kt
```diff
@@ -6,13 +6,17 @@
...
+@file:Suppress("DEPRECATION")
+
 package org.readium.r2.shared.publication.presentation
 
 import org.json.JSONObject
 import org.readium.r2.shared.publication.Metadata
 
 // Presentation extensions for [Metadata]
 
+@Deprecated("This was removed from RWPM. You can still use the EPUB extensibility to access the original values.")
```
mickael-menu: Maybe use `level = ERROR` here? Since presentation metadata is no longer populated when parsing an EPUB, the property is effectively broken.

qnga: Good point

mickael-menu (approved):
Thank you for making the changes

PR #767, opened 2026-03-20
Animated page turns

### LINKED ISSUE #766: goForward(true) not animating page turns
### Describe the bug

Is `EpubNavigatorFragment.goForward(true)` supposed to be animating page turns? Because for me it is only animating the transition between sections/chapters, but page turns are always instantaneous.

https://github.com/user-attachments/assets/1e5fcf3b-2a95-4616-a10f-ba2feed1788a

### How to reproduce?

Open a book and perform an action that will programmatically change pages with `EpubNavigatorFragment.goForward(true)` (in my case, it is a tap handler for the right side of the page).

### Readium version

3.1.2

### Android API version

35

### Additional context

_No response_

marcprux:
Implements support for animated page turns. Closes https://github.com/readium/kotlin-toolkit/issues/766.

Leaving in draft state until I get a chance to test thoroughly.


marcprux:
It works pretty well for me.

https://github.com/user-attachments/assets/aea9c008-b957-46db-9d8d-b901ef4dd77f

stevenzeck:
Does this just add animations to page turns when tapping the edges in the same way that swiping does?

marcprux:
> Does this just add animations to page turns when tapping the edges in the same way that swiping does?

Yes, whatever action might trigger programmatic calls to `EpubNavigatorFragment.goForward(true)` will now cause the page turn to animate in a way that (mostly) matches the existing animation that already happens when navigating between sections.

This matches the behavior of iOS, except that iOS currently _doesn't_ animate the jump between sections, only page turns within sections.

mickael-menu (commented):
Thank you for tackling that Marc!

Did you consider using `scrollTo()` JS API with `smooth` behavior? I found the UX better when tapping rapidly on the edge of the screen without, but we can't control the duration of the animation AFAIK, which is on the slower side. This is what I used on the Swift toolkit.

Here's a patch from your branch if you want to try it out: [scroll-to.patch](https://github.com/user-attachments/files/26246604/scroll-to.patch)

Would you mind updating the CHANGELOG, or enabling Edit from maintainers on your PR?

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index b9e9bb72d..12d78d243 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file. Take a look

 #### Navigator

+* Support for animated page turns in the EPUB navigator (contributed by [@marcprux](https://github.com/readium/kotlin-toolkit/pull/767)).
 * New experimental positioning of EPUB decorations that places highlights behind text to improve legibility with opaque decorations (contributed by [@ddfreiling](https://github.com/readium/kotlin-toolkit/pull/721)).
     * To opt-in, initialize the `EpubNavigatorFragment.Configuration` object with `decorationTemplates = HtmlDecorationTemplates.defaultTemplates(alpha = 1.0, experimentalPositioning = true)`.
```

marcprux:
> Did you consider using `scrollTo()` JS API with `smooth` behavior?

I experimented with it briefly and wasn't entirely satisfied with the feel and the speed. I agree it is much simpler, though, so I'd be fine with just switching to using that.

> Here's a patch from your branch if you want to try it out: [scroll-to.patch](https://github.com/user-attachments/files/26246604/scroll-to.patch)

That is `Not Found` for me, but I expect I know the contents.

> or enabling Edit from maintainers on your PR?

I would but I [can't](https://github.com/orgs/community/discussions/5634) due to the fork being under the Stanza-Redux org. But if you get me your proposed patch I'll integrate it and update the CHANGELOG. Or feel free to throw away this PR and start a new one, which might just be simpler 😄

mickael-menu:
Thank you @marcprux, I ended up keeping your solution because the difference was too jarring between columns and resource transitions. I modified it a bit to handle rapid tap edges.

Superseded by https://github.com/readium/kotlin-toolkit/pull/789

PR #776, opened 2026-04-19
Update jsoup

stevenzeck:
Update jsoup and its associated tests. The breaking changes from the tests are from:

1. https://github.com/jhy/jsoup/pull/2283. jsoup prefers shorter selectors using the nearest unique ancestor ID.
2. Only void elements can have self-closing tags, which `<audio>` and `<video>` are not (hence why the test needed an update). I have to look through what are considered void elements, but this could potentially cause issues depending on the HTML in the epub. https://github.com/jhy/jsoup/issues/2300

For reference, void elements are:

- `<area>`
- `<base>`
- `<br>`
- `<col>`
- `<embed>`
- `<hr>`
- ``
- `<input>`
- `<link>`
- `<meta>`
- `<param>`
- `<source>`
- `<track>`
- `<wbr>`

Still needs to be tested thoroughly.


gradle/libs.versions.toml
```diff
@@ -40,7 +40,7 @@ google-material = "1.13.0"
 joda-time = "2.14.0"
 #Do not upgrade without fixing unit tests in HtmlResourceContentIteratorTest and regression testing
 #See https://github.com/jhy/jsoup/pull/2283
```
mickael-menu, line 42: You should be able to remove these comments, right?

stevenzeck: Yeah, I'll do that once the PR is in review status so it triggers checks.

stevenzeck:
Did some testing of a reflowable in the testapp, no issues stood out. I still want to test fixed format as well as the navigator demo.

My only concern is the second point in the original comment: Are there elements in epubs that are void but are commonly marked up as self-closing?

mickael-menu:
EPUBs are supposed to be XHTML, and there are no custom EPUB tags. However this code needs to support Web Publications which might be HTML 4 or 5. What happens with `<audio/>` using Jsoup?

AFAIK `<audio/>` is indeed incorrect, in both XHTML and HTML.

stevenzeck:
In practice, I don't know. In the tests, the AudioElement's Locator's otherLocations comes out as `html > body > audio > audio` instead of `html > body > audio:nth-child(2)`. I haven't dug into the exact way jsoup parses it. Their relevent issue is https://github.com/jhy/jsoup/issues/2300.

mickael-menu:
Okay, I think this should not break anything.

stevenzeck:
Everything looks ok to me.

gradle/libs.versions.toml
mickael-menu: You can remove the comments now.

stevenzeck: Done

mickael-menu (approved):
Thank you!

PR #778, opened 2026-04-24
Remove material icons extended dependency

stevenzeck:
Google is no longer supporting the material icon library, instead it's back to using the Google Fonts' XML and importing. That dependency is 20 MB anyways, and we don't need it all.


qnga (commented):
Why did you need to change the tint of many icons?

stevenzeck:
> Why did you need to change the tint of many icons?

Force of habit. It's used for light/dark mode and material you.

qnga (approved):
Thanks! I couldn't find any difference without tints. Let's keep things simple.

PR #792, opened 2026-05-16
Fix Dokka

stevenzeck:
The recent updates to Dokka and AGP caused the kdoc documentation generation to break. This fixes it.


qnga (approved):
Sounds good, thanks.

PR #794, opened 2026-05-21
Replace NetworkService with HttpClient in LCP

stevenzeck:
This PR replaces the usage of NetworkService with HttpClient in the LCP module.

Related to #779.


readium/lcp/src/main/java/org/readium/r2/lcp/license/License.kt
```diff
@@ -185,11 +188,14 @@ internal class License private constructor(
                 parameters["end"] = endDate.toString()
             }
 
-            val url = link.url(parameters = parameters)
+            val url = link.url(parameters = parameters) as? AbsoluteUrl
+                ?: throw LcpException(LcpError.Parsing.Url(link.rels.firstOrNull() ?: ""))
```
qnga:
```suggestion
                ?: throw LcpException(LcpError.Parsing.Url(link.rels.first()))
```
Rel is mandatory and the parsing fails if it's missing.

readium/lcp/src/main/java/org/readium/r2/lcp/service/LicensesService.kt
```diff
@@ -152,16 +156,20 @@ internal class LicensesService(
         onProgress: (Double) -> Unit,
     ): LcpService.AcquiredPublication {
         val link = license.link(LicenseDocument.Rel.Publication)!!
-        val url = link.url()
+        val url = link.url() as? AbsoluteUrl
+            ?: throw LcpException(LcpError.Parsing.Url(link.rels.firstOrNull() ?: ""))
```
qnga: Same as above.

readium/lcp/src/test/java/org/readium/r2/lcp/service/CRLServiceTest.kt
```diff
@@ -7,22 +7,47 @@
...
+                        statusCode = HttpStatus.Success,
+                        headers = emptyMap(),
+                        mediaType = null
+                    ),
+                    ByteArrayInputStream(ByteArray(0))
+                )
+            )
+        }
+    }
+
```
qnga, line 50: No more mockk, that's great.

readium/shared/src/main/java/org/readium/r2/shared/util/http/HttpClient.kt
```diff
@@ -145,3 +150,71 @@ public suspend fun HttpClient.head(request: HttpRequest): HttpTry<HttpResponse>
...
+                                        }
+                                    }
+                                    lastProgress = progress
+                                }
+                            }
+                        }
+                    }
+                }
+                Try.success(
+                    HttpFetchResponse(response.response, ByteArray(0))
```
qnga: I don't understand why you return an empty ByteArray. You could return nothing. Besides, the error handling is a bit simplistic. That might be perfectionism, but in case of a filesystem error, we shouldn't fail with an `HttpError`. I would rather introduce an `HttpDownloadError` which could be either an `HttpError` or a `FilesystemError`.

readium/lcp/src/main/java/org/readium/r2/lcp/service/LicensesService.kt
```diff
@@ -157,19 +158,23 @@ internal class LicensesService(
...
         val response = httpClient.download(
             HttpRequest(url),
             destination,
             onProgress = onProgress
-        ).getOrElse {
-            throw LcpException(LcpError.Network(Exception(it.message)))
+        ).getOrElse { error ->
+            when (error) {
+                is HttpDownloadError.Http -> throw LcpException(LcpError.Network(cause = error.cause))
+
```
qnga, line 172: I think there is a fallback to the link media type in `NetworkService`.

readium/shared/src/main/java/org/readium/r2/shared/util/http/HttpClient.kt
stevenzeck: I added a `HttpDownloadError` sealed class in `HttpError` and followed the naming pattern in the latter: `HttpDownloadError.Http` and `HttpDownloadError.Filesystem`. I'm wary about naming conventions, as we already have an `HttpError` class. And there are a lot of files and classes already with the same name, which gets confusing and causes conflicts that have to be addressed.

Could also make `HttpDownloadError` an interface that HttpError implements, but that might be crossing boundaries a bit too much.

qnga: > And there are a lot of files and classes already with the same name, which gets confusing and causes conflicts that have to be addressed.

Could you elaborate on this?

readium/lcp/src/main/java/org/readium/r2/lcp/service/LicensesService.kt
stevenzeck: Yes, that was already handled in here a few lines below: `license.publicationLink.mediaType ?: serverMediaType`. NetworkService had an additional check which wasn't really necessary, and I didn't want to include all of that logic in the download extension function.

qnga: Ok, I missed that!

readium/shared/src/main/java/org/readium/r2/shared/util/http/HttpClient.kt
stevenzeck: It's nothing specific to this PR. One example is Encryption in [Shared](https://github.com/readium/kotlin-toolkit/blob/6d89b0f508f448a64285f1a4189c3ccb32d94d90/readium/shared/src/main/java/org/readium/r2/shared/publication/encryption/Encryption.kt) and in [LCP](https://github.com/readium/kotlin-toolkit/blob/6d89b0f508f448a64285f1a4189c3ccb32d94d90/readium/lcp/src/main/java/org/readium/r2/lcp/license/model/components/lcp/Encryption.kt). There's a similar one with `Link`.

I think most of it is older stuff, but don't want to add to it.

qnga: Ok! I think it's fine to have different types as they conform to different specifications. Even if they currently match, they might diverge at some point.

stevenzeck: Yeah having different types as fine, I'm more against the same naming.

readium/lcp/src/main/java/org/readium/r2/lcp/service/CRLService.kt
```diff
@@ -54,7 +58,9 @@ internal class CRLService(val network: NetworkService, val context: Context) {
 
     private suspend fun fetch(): String {
         val url = "http://crl.edrlab.telesec.de/rl/EDRLab_CA.crl"
-        val data = network.fetch(url, NetworkService.Method.GET)
+        val absoluteUrl = AbsoluteUrl(url) ?: throw LcpException(LcpError.CrlFetching)
```
qnga: Looks like my comment was lost. "http://crl.edrlab.telesec.de/rl/EDRLab_CA.crl" is a well-known absolute Url. You can use `AbsoluteUrl("")!!`.

readium/shared/src/main/java/org/readium/r2/shared/util/http/HttpClient.kt
```diff
@@ -145,3 +150,69 @@ public suspend fun HttpClient.head(request: HttpRequest): HttpTry<HttpResponse>
...
+                            }
+                        }
+                    }
+                }
+                Try.success(success = response.response)
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: IOException) {
+                Try.failure(
+                    failure = HttpDownloadError.Filesystem(exception = e)
```
qnga: That's not correct. `IOException`s can be either filesystem or network errors. You need to catch them more locally to be sure.

readium/shared/src/main/java/org/readium/r2/shared/util/http/HttpError.kt
```diff
@@ -74,3 +74,24 @@ public sealed class HttpError(
...
+public sealed class HttpDownloadError(
+    public override val message: String,
+    public override val cause: Error? = null,
+) : Error {
+
+    /** An HTTP error occurred during the download. */
+    public class Http(error: HttpError) : HttpDownloadError(message = error.message, cause = error)
+
+    /** A filesystem error occurred while saving the downloaded resource. */
+    public class Filesystem(cause: Error) : HttpDownloadError(
```
qnga: You could use the `FileSystemError` type for the property. You can map `SecurityException`s to `Forbidden` and `IOException`s to `IO`.

qnga:
I tried to simplify and clarify `HttpClient.download`. Here are some explanations:
- Cancellation is cooperative. It cannot happen in synchronous code when you don't check for cancellation.
- I don't understand why you cared about rounding. Let's keep it simple if there is no important reason.
- In case of failure, the file must be deleted. It could be left to the caller, but I don't think you dealt with this there either.

readium/lcp/src/main/java/org/readium/r2/lcp/LcpError.kt
```diff
@@ -242,8 +241,8 @@ public sealed class LcpError(
 
         internal fun wrap(e: Exception): LcpError = when (e) {
             is LcpException -> e.error
-            is NetworkException -> Network(e)
             is SocketTimeoutException -> Network(e)
+            is java.io.IOException -> Network(e)
```
qnga: Again a comment not published. Is this meant for something in particular? I don't think that's correct.

stevenzeck: It was leftover from what I was originally doing to replace the old NetworkException.

qnga (approved):
Looks good, thanks.

PR #795, opened 2026-05-23
Add the `scroll` preference to the PDFium adapter

ddfreiling:
Adds preference to PDFium to support a paginated scroll feel (e.g. `scroll=false`).
This aligns `kotlin-toolkit` further with the PDF features of `swift-toolkit`.

Where `swift-toolkit` defaults to horizontal pagination, `kotlin-toolkit` with pdfium adapter defaults to vertical continuous scrolling. I'm not sure if these defaults are worth aligning, but let me know if that's something you want.


qnga:
I don't think this is what a paginated mode is supposed to be or what we've got in the Swift toolkit. It does snap on pages but they are not clearly separated. Do you want to offer this in production? I'm not sure if it's worth it. @mickael-menu, an opinion?

readium/adapters/pdfium/navigator/src/main/java/org/readium/adapter/pdfium/navigator/PdfiumPreferences.kt
```diff
@@ -18,13 +18,16 @@ import org.readium.r2.navigator.preferences.ReadingProgression
...
  *  @param scrollAxis Indicates the axis along which pages should be laid out in scroll mode.
  */
 @Serializable
 public data class PdfiumPreferences(
     val fit: Fit? = null,
     val pageSpacing: Double? = null,
     val readingProgression: ReadingProgression? = null,
+    val scroll: Boolean? = null,
     val scrollAxis: Axis? = null,
 ) : Configurable.Preferences<PdfiumPreferences> {
```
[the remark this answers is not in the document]
mickael-menu: I understand the point raised by Copilot, but I actually prefer your initial implementation. We keep the preferences alphabetically sorted for consistency. People should always use named arguments for such objects with long property lists.

CHANGELOG.md
```diff
@@ -4,7 +4,13 @@ All notable changes to this project will be documented in this file. Take a look
...
 **Warning:** Features marked as *experimental* may change or be removed in a future release without notice. Use with caution.
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Added
+
+#### Navigator
+
+* Added a `scroll` preference to the PDF navigator (PDFium adapter) to match the Swift toolkit. When set to `false`, page flings snap to page boundaries for a paginated feel. Defaults to `true` (current continuous-scroll behavior).
```
mickael-menu: We could keep the changelog focused on the changes within the Kotlin toolkit and leave out the mention of the Swift toolkit.

docs/guides/navigator/preferences.md
```diff
@@ -466,3 +466,7 @@ Some preferences are available only in scroll or paginated mode (`scroll = false
 | `spread`          |                    | :white_check_mark: |
 | `scrollAxis`      | :white_check_mark: |                    |
 
+### PDF (PDFium)
+
+The PDFium adapter exposes a `scroll` preference for parity with the Swift toolkit. When set to `false`, page flings snap to page boundaries (paginated feel). Defaults to `true` (continuous scrolling).
```
mickael-menu: Given my other main comment, this could be updated with a table similar to the PSPDFKit entry.

mickael-menu (commented):
Thanks, Daniel! I think this is a valuable addition that aligns well with what we have in the Swift toolkit. The only minor issue, which is likely what @qnga was referring to, is that when `scroll` is set to `true`, the `scrollAxis` should probably be disabled and fixed to `horizontal`.

- With PSPDFKit: https://github.com/readium/kotlin-toolkit/blob/523c5873487007662db64302364628879f9e6c22/readium/adapters/pspdfkit/navigator/src/main/java/org/readium/adapter/pspdfkit/navigator/PsPdfKitPreferencesEditor.kt#L134
- And in Swift: https://github.com/readium/swift-toolkit/blob/5447aaffe6723254f9566c78fe6364232d9a3276/Sources/Navigator/PDF/Preferences/PDFPreferencesEditor.swift#L107

> Where `swift-toolkit` defaults to horizontal pagination, `kotlin-toolkit` with pdfium adapter defaults to vertical continuous scrolling. I'm not sure if these defaults are worth aligning, but let me know if that's something you want.

I agree that we should align this with the Swift toolkit as well. Hadrien and I discussed this a while ago, and since the toolkit primarily focuses on e-book reading apps, we decided to set `scroll = false` as the default.

qnga:
That's far better in horizontal mode, indeed. I'm not sure why the paginated mode had never been implemented if it's supported by the rendering engine.

ddfreiling:
For some reason I forgot all about making this PR, will follow-up tomorrow.

ddfreiling:
@mickael-menu should now be updated according to review. Was there a decision on default `scroll` direction, where the toolkits currently differs?

ddfreiling:
@mickael-menu Any preference on aligning the default scroll direction?

mickael-menu:
Sorry I missed your comment. The default scroll direction should be horizontal and paginated, as we're a reading toolkit first and not a generic PDF reader. I'll try to review and get this PR merged this Wednesday, thank you.

mickael-menu (approved):
Thank you for making the changes @ddfreiling 🙏

mickael-menu:
Ha I can't push to your fork. I opened another PR with a couple fixes: https://github.com/readium/kotlin-toolkit/pull/825

ddfreiling:
Thanks @mickael-menu , appreciate it 🚀

PR #797, opened 2026-06-01
Allow assets to be consumed in publication resources

qnga:


mickael-menu (approved):
Thank you, looking good to me.

PR #760, opened 2026-03-01
Fix 3 crashes

RankoR:
These 3 commits fix 3 crashes occurring in production:

## Invocations on zero-size WebView:

```
          Fatal Exception: java.lang.IllegalStateException: Check failed.
       at org.readium.navigator.web.internals.webview.WebViewScrollController.moveToProgression(WebViewScrollController.kt:111)
       at org.readium.navigator.web.reflowable.resource.ReflowableResourceKt$ReflowableResource$1$3$1.invokeSuspend$lambda$0$0$0$0(ReflowableResource.kt:200)
       at org.readium.navigator.web.reflowable.resource.ReflowableResourceKt$ReflowableResource$1$3$1.$r8$lambda$hRiAUJQolcOVqg3AKXicl-UOMLY()
       at org.readium.navigator.web.reflowable.resource.ReflowableResourceKt$ReflowableResource$1$3$1$$ExternalSyntheticLambda2.invoke(D8$$SyntheticClass)
       at org.readium.navigator.web.internals.webview.WebViewUtilKt$invokeOnReadyToBeDrawn$1$1.onComplete(WebViewUtil.kt:21)
       at WV.w13.a(chromium-SystemWebViewGoogle6432.aab-stable-763207903:3)
       at WV.yf.run(chromium-SystemWebViewGoogle6432.aab-stable-763207903:7)
       at android.os.Handler.handleCallback(Handler.java:995)
       at android.os.Handler.dispatchMessage(Handler.java:103)
       at android.os.Looper.loopOnce(Looper.java:273)
       at android.os.Looper.loop(Looper.java:363)
       at android.app.ActivityThread.main(ActivityThread.java:10060)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:632)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)
        
```

## Unwrapping null value

```
          Fatal Exception: java.lang.NullPointerException:
       at org.readium.navigator.web.internals.pager.RenditionScrollState.onDocumentResized(RenditionScrollState.kt:186)
       at org.readium.navigator.web.reflowable.ReflowableWebRenditionKt.ReflowableWebRendition$lambda$0$0$1$6$0(ReflowableWebRendition.kt:192)
[...]

PR #799, opened 2026-06-05
Fix crashes on zero-size WebView and disposed scroll controller

qnga:
Try to mitigate issues reported in #760.


mickael-menu (approved):
Looking good to me, thanks

PR #801, opened 2026-06-10
Fix Dokka

stevenzeck:
Dokka has a scripts folder now, and the mkdocs file was excluding all folders named `scripts/`. We only want to exclude the root scripts folder, so changed it to `/scripts/`.


mickael-menu (approved):
Thanks

PR #802, opened 2026-06-17
Update Android SDK and other dependencies

stevenzeck:
Leaving in draft until Kotlin Multiplatform supports Android Gradle Plugin 9.1.1, since SDK 37 requires that.

### Changes

1. Android target and compile SDK set to 37.
2. Updates Kotlin to 2.4.20.
3. Updates Android Gradle Plugin to 9.3.1.
4. Updates room to room3.
5. Updates other dependencies.
6. Addresses deprecations from said updates.
7. Updates code to use backing fields introduced in Kotlin 2.4.0.

### Notes

1. Android Studio will show errors for the backing fields. It's a known issue, but harmless and builds fine. You can add `freeCompilerArgs.add("-XXLanguage:+ExplicitBackingFields")` to the kotlin > compilerOptions block in gradle to suppress it.
2. Need to test LCP passphrases that are already saved prior to this update to ensure no issues with room to room3.


stevenzeck:
Going to split this up into a few PRs.

stevenzeck:
On second thought, things are too intertwined, especially with newer Jetpack releases requiring minSdk 24 and/or AGP versions.

buildSrc/src/main/kotlin/readium.library-conventions.gradle.kts
```diff
@@ -103,6 +100,6 @@ mavenPublishing {
         }
     }
 
-    publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL)
+    publishToMavenCentral()
```
mickael-menu, line 103: Why did you change this?

.github/workflows/publish.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           java-version: 17
 
       - name: Publish to Maven Central
-        run: ./gradlew publishToMavenCentral --no-configuration-cache
+        run: ./gradlew publishToMavenCentral
```
mickael-menu, line 22: Why did you change this?

readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2EpubPageFragment.kt
```diff
@@ -170,6 +179,12 @@ internal class R2EpubPageFragment : Fragment() {
         webView.settings.builtInZoomControls = true
         webView.settings.displayZoomControls = false
         webView.settings.textZoom = textZoom
+        if (WebViewFeature.isFeatureSupported(WebViewFeature.DOWNLOAD_FAVICONS_ENABLED)) {
+            WebSettingsCompat.setDownloadFaviconsEnabled(
+                webView.settings,
+                false
+            )
+        }
```
mickael-menu: This is duplicated three times. We could extract it into a `WebSettings.disableFavicons()` extension.

Also this, and the `onRenderProcessGone()` seem unrelated to updating dependencies, maybe they could be in their own PRs? What motivated these changes?

readium/lcp/build.gradle.kts
```diff
@@ -15,12 +16,16 @@ android {
...
 kotlin {
     compilerOptions {
-        languageVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_3
+        languageVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_4
         // See https://github.com/readium/kotlin-toolkit/pull/525#issuecomment-2300084041
         freeCompilerArgs.add("-Xconsistent-data-class-copy-visibility")
     }
 }
 
+room3 {
```
mickael-menu: It looks like we can remove this, since we're using `exportSchema = false`.

readium/adapters/pdfium/navigator/src/test/java/org/readium/adapter/pdfium/navigator/LegacyLocatorMigrationTest.kt
```diff
@@ -23,9 +23,11 @@ import org.readium.r2.shared.publication.services.positionsServiceFactory
 import org.readium.r2.shared.util.Url
 import org.readium.r2.shared.util.mediatype.MediaType
 import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
 
 @OptIn(DelicateReadiumApi::class)
 @RunWith(RobolectricTestRunner::class)
+@Config(sdk = [36])
```
mickael-menu: Do we need to specify an explicit SDK version here? Since the SDK is already defined in `src/test/resources/robolectric.properties`, perhaps we could use that instead?

readium/lcp/build.gradle.kts
```diff
@@ -33,8 +38,8 @@ dependencies {
     implementation(libs.androidx.browser)
     implementation(libs.kotlinx.datetime)
 
-    implementation(libs.bundles.room)
-    ksp(libs.androidx.room.compiler)
+    implementation(libs.androidx.room3)
```
mickael-menu, line 36: Note to self: verify the migration from old LCP databases.

test-app/build.gradle.kts
```diff
@@ -61,11 +62,15 @@ android {
 
 kotlin {
     compilerOptions {
-        languageVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_3
-        freeCompilerArgs.add("-Xannotation-default-target=param-property")
+        languageVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_4
+        freeCompilerArgs.add("-XXLanguage:+ExplicitBackingFields")
```
mickael-menu: Would you happen to have a reference to share for this known issue? Also, would it make sense to move this into the toolkit instead of keeping it in the test app's `build.gradle.kts`?

Personally I didn't see any errors or warnings after commenting this out, with `./gradlew clean build` and the latest Android Studio.

buildSrc/src/main/kotlin/readium.library-conventions.gradle.kts
stevenzeck: Per the [0.33.0](https://github.com/vanniktech/gradle-maven-publish-plugin/releases/tag/0.33.0) release notes:

- publishToMavenCentral() and publishToMavenCentral(automaticRelease) without SonatypeHost will now publish through the Central Portal.
- Deprecated overloads of publishToMavenCentral that take a SonatypeHost parameter.

.github/workflows/publish.yml
stevenzeck: [0.34.0](https://github.com/vanniktech/gradle-maven-publish-plugin/releases/tag/0.34.0) added support for configuration cache. I thought that was there because it wasn't compatible with publishToMavenCentral at the time. Can remove it if you want.

readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2EpubPageFragment.kt
stevenzeck: They are both new features with webkit [1.17.0](https://developer.android.com/jetpack/androidx/releases/webkit#webkit_version_117_2). I didn't want to just update the version without using any useful new features. I can pull them out of this into their own PR if you want.

.github/workflows/publish.yml
mickael-menu: It might be, from: https://kotlinlang.org/docs/multiplatform/multiplatform-publish-libraries-to-maven.html#add-a-github-actions-workflow-to-your-project

Image

The linked issue is still marked as open, but the changelog you shared seems to contradict it. Let's try without and see if it works during the next release.

buildSrc/src/main/kotlin/readium.library-conventions.gradle.kts
mickael-menu: Thank you

test-app/build.gradle.kts
stevenzeck: It was an issue in an older Android Studio build. Yeah, I'll remove it now, doesn't seem to be a problem.

readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2EpubPageFragment.kt
mickael-menu: No that's fine, maybe just extract the helper?

I have some doubts about how to handle `onRenderProcessGone()` though. I think I would prefer to remove it for now and reintroduce it if we encounter an issue that it can actually solve.

readium/adapters/pdfium/navigator/src/test/java/org/readium/adapter/pdfium/navigator/LegacyLocatorMigrationTest.kt
stevenzeck: It was only this specific test class that was failing. I can create a robolectric.properties file for pdfium-navigator if you want, it just didn't seem worth it for one file.

mickael-menu: I've seen this change in the lcp module as well in your PR. But yes, I think it's best to stay consistent throughout the toolkit and use the `robolectric.properties` file.

readium/lcp/build.gradle.kts
stevenzeck: We actually don't need the room gradle plugin, I don't know why I put it in there. That block is required when using it.

.github/workflows/publish.yml
stevenzeck: From what I read in that issue, it was because of publishing to the older Sonatype Nexus. Since they no longer publish there, the blocker from https://github.com/gradle/gradle/issues/22779 no longer applied.

readium/navigator/src/main/java/org/readium/r2/navigator/pager/R2EpubPageFragment.kt
stevenzeck: I'll just remove everything so we don't have to deal with visibility in the code. We can adopt these in the future as we see fit.

mickael-menu: Sounds good

mickael-menu (approved):
Thank you @stevenzeck

PR #803, opened 2026-06-19
Add `targetElement` to pointer events

mickael-menu:
#### Navigator

* The EPUB navigator now reports the image under a tap through the new experimental `TapEvent.targetElement` property, making it easy to build features such as image previews. See the [EPUB Image Preview guide](docs/guides/navigator/epub-image-preview.md).

---

This implementation was automatically ported from https://github.com/readium/swift-toolkit/pull/747 using the following plan:

<details>
  <summary>Port "Support Image Previews" readium/swift-toolkit#747 to Kotlin</summary>

# Port "Support Image Previews" (swift-toolkit #747) to Kotlin

## Context

The Swift toolkit PR [[#747](https://github.com/readium/swift-toolkit/pull/747)](https://github.com/readium/swift-toolkit/pull/747) lets apps
detect when a user taps an image (``, `<svg>`) in an EPUB and obtain its metadata
(manifest link, caption, accessibility label, CSS-selector locator, and on-screen frame),
without writing custom gesture/hit-testing code. The motivating use case is image zoom
(readium/mobile#10). This ports the same capability to the Kotlin toolkit.

**Key architectural difference vs. Swift.** Swift already had a unified `PointerEvent` with a
coordinate-conversion pipeline in `EPUBSpreadView`, so the PR just enriched it. Kotlin has
**no** `PointerEvent`. Taps flow through two *separate* navigator stacks, each with its own
gesture script and `TapEvent` type:

- **Legacy WebView navigator** (`EpubNavigatorFragment`, used by the test app):
  JS `onClick` → `Android.onTap(json)` → `R2BasicWebView.onTap` → `Listener.onTap(point)`
  → `inputListener.onTap(TapEvent(point))`. Public type
  `org.readium.r2.navigator.input.TapEvent(point: PointF)`. JS currently sends `targetElement`
  as raw `outerHTML`. Coordinates are `clientX * devicePixelRatio`.
- **Compose `web` navigator** (preview): TypeScript `gestures.ts` → `all-listener-bridge.ts`
  → `GesturesApi.onTap(json)` → `GesturesListener.onTap(DpOffset)` → `DelegatingGesturesListener`
  → `org.readium.navigator.common.TapEvent(offset: DpOffset)` →
  `InputListener.onTap(event, TapContext)`. Works in **dp** (no pixel-ratio scaling).

So this is an **adaptation**, not a transcription.

### Decisions (confirmed with user)
- **API surface:** add `targetElement` to each navigator's existing public `TapEvent` (closest
  analog to Swift's `PointerEvent.targetElement`), gated `@ExperimentalReadiumApi`.
- **Linked images out of scope (matches Swift):** images wrapped in interactive elements
  (`<a>`) are filtered by the interactive-element check before reaching `onTap`. Swift's
  PR has the same limitation (its `didReceivePointerEvent` guards `interactiveElement == nil`).
  We keep this behavior and document it as a known limitation; do **not** modify the
  interactive-element filter.
- **Sequencing:** Phase 1 = legacy **reflowable** navigator + shared model + test app + docs
  (complete, demonstrable). Phase 2 = Compose `web` navigator (follow-up), which also covers
  fixed-layout.
- **Legacy fixed-layout out of scope.** Legacy FXL taps are detected *natively* by
  `R2FXLLayout` and call `listener.onTap(PointF(...))` directly (`R2FXLPageFragment.kt:91/115`)
  — they never pass through `gestures.js`/`Android.onTap`, so no element metadata is available
  without separate work. (`currentReflowablePageFragment` is also `R2EpubPageFragment`-only,
  i.e. null for FXL.) Fixed-layout image preview is delivered by the Compose `web` navigator in
  Phase 2 (the modern FXL path). The legacy FXL `onTap` call sites simply pass
  `targetElement = null` after the signature change.
- **Link resolution:** resolved natively. Legacy looks up the manifest `Link`; web synthesizes
  a `Link` (see Phase 2 constraint).
- **SVG:** full parity — add a new shared `Content.SvgElement` for inline `<svg>`.
- **Frame:** include the element's on-screen rect, in **each navigator's own coordinate
  convention** (RectF in device px for legacy; DpRect for web).
- **Docs:** CHANGELOG entry + a Kotlin "EPUB Image Preview" guide.
- **Test app:** simple full-screen dialog showing the tapped image + caption.

## Shared model (Phase 1)

`readium/shared/.../publication/services/content/Content.kt`
- Add `Content.SvgElement` next to `ImageElement` (Swift `SVGContentElement`):
  ```kotlin
  public data class SvgElement(
      override val locator: Locator,
      val svg: String,            // raw inline SVG source
      val caption: String?,
      override val attributes: List<Attribute<*>> = emptyList(),
  ) : TextualElement {
      override val text: String? get() = caption?.takeIf { it.isNotBlank() } ?: super.text
  }
  ```
  `text` falls back caption → `super.text` (accessibility label), identical to `ImageElement`
  (both read `accessibilityLabel` via `AttributesHolder`). `ImageElement` already matches
  Swift `ImageContentElement`.

**Element-type discrimination (mirror Swift exactly — by resolvable source, not by tag):**
`findNearestImageElement` matches `` and `<svg>`. The routing rule:
- A resolvable `src`/`href` was found → build an `embeddedLink` → **`Content.ImageElement`**.
  This covers `` (incl. ``) and the rare `<svg href=...>`.
- Otherwise, an **inline** `<svg>…</svg>` (no resolvable source) → **`Content.SvgElement`** with
  the raw `outerHTML`.
- If neither applies (e.g. an `<svg>` with no source *and* no usable html) → return `null`
  (no `targetElement`). Note: `<svg src=...>` is **not** a real construct — discrimination is on
  source presence, not the tag name.

`readium/shared/.../publication/Locator.kt`
- CSS selector lives in `Locations.otherLocations["cssSelector"]` (no schema change). Add a
  convenience accessor for parity with Swift's `cssSelector` setter:
  `public val Locator.Locations.cssSelector: String? get() = otherLocations["cssSelector"] as? String`.
  Build locators with `Locations(otherLocations = mapOf("cssSelector" to selector))`.

### JavaScript — `src/main/assets/_scripts/src/`
- In `gestures.js`, port the Swift helpers `extractTargetElement`, `extractCaption`,
  `findNearestImageElement` (verbatim logic). Resolve `src` via
  `new URL(rawSrc, document.baseURI).href`. Use `toNativeRect()` from `rect.js` for the frame
  (multiplies by `devicePixelRatio`, matching the legacy `clientX * devicePixelRatio` tap
  coords).
- Import `getCssSelector` from `css-selector-generator` (already a dependency).
- The existing `clickEvent.targetElement = event.target.outerHTML` string is **dead** (no JS or
  Kotlin reader; `handleFootnote` uses `interactiveElement`, and the native
  `TapEvent.targetElement: String` at `R2BasicWebView.kt:333/346` is parsed but never read).
  **Replace** it with the structured metadata object — no collision handling needed.
- **Decoration path shares this `clickEvent`.** `handleDecorationClickEvent(event, clickEvent)`
  forwards the whole `clickEvent` (now with the structured `targetElement`) as `click` to
  `Android.onDecorationActivated`, which parses it via the shared `TapEvent.fromJSONObject` and
  reads only `.point`. `decorator.js` does not read `targetElement`. So the shape change is
  benign, but the shared parser must tolerate `targetElement` being an object/absent. Keep the
  parser defensive and verify decoration activation still works (see verification).
- **Performance:** extraction is already gated — `extractTargetElement` early-returns at
  `findNearestImageElement` *before* `getBoundingClientRect` / `new URL` / `getCssSelector`, so
  ordinary non-image taps (UI toggle, navigation) only pay a cheap parent-walk. No extra gating
  needed.
- Rebuild: `cd src/main/assets/_scripts && pnpm install && pnpm run bundle`.

### Native — Kotlin
`R2BasicWebView.kt`
- Replace the dead `targetElement: String` field in the private `TapEvent` with a parsed
  holder (`tag, html, src, frame: RectF, accessibilityLabel, caption, cssSelector`), parsed in
  `fromJSONObject`.
- Extend `Listener.onTap` to carry it:
  `fun onTap(point: PointF, targetElement: TargetElementData?): Boolean`. Update call sites:
  `R2BasicWebView.onTap` (passes the parsed metadata) and the two legacy-FXL sites in
  `R2FXLPageFragment.kt:91/115` (pass `targetElement = null`, since FXL bypasses gestures.js).
  Linked-image filtering is unchanged: a non-null
  `interactiveElement` still short-circuits to `handleFootnote` before `onTap`.

`input/TapEvent.kt`
```kotlin
public data class TapEvent(
    val point: PointF,
    @ExperimentalReadiumApi val targetElement: TargetElement? = null,
)
@ExperimentalReadiumApi
public data class TargetElement(val frame: RectF, val content: Content.Element)
```

`epub/EpubNavigatorFragment.kt` (`WebViewListener.onTap`)
- Build the public `TargetElement`:
  - Resolve the resource manifest `Link` from the reflowable page fragment owning the WebView
    (`currentReflowablePageFragment?.link`). Build the locator
    `Locator(href = link.url(), mediaType = ..., locations = Locations(otherLocations =
    mapOf("cssSelector" to selector)))`.
  - Relativize the JS `src` against the publication base URL with
    `Url(src)?.let { baseUrl.relativize(it) }`, then `publication.linkWithHref(relativized)`
    (fallback to synthesised `Link(href)` for external/unmanifested URLs). See
    `Manifest.linkWithHref`, `Url.relativize`.
  - Map to `Content.ImageElement` / `Content.SvgElement` per the discrimination rule above.
    Attach `accessibilityLabel` as `Content.Attribute(AttributeKey.ACCESSIBILITY_LABEL, label)`.
  - Apply `frame.adjustedToViewport()` (existing helper) for the paddingTop offset.
- Pass it into `inputListener.onTap(TapEvent(point, targetElement))`.

## Phase 1 — Test app (`test-app`)

`reader/VisualReaderFragment.kt`
- Add an `InputListener` **before** the existing toggle-system-UI listener (lines 124–131).
  The example must include `@OptIn(ExperimentalReadiumApi::class)` to surface the gated field:
  ```kotlin
  @OptIn(ExperimentalReadiumApi::class)
  override fun onTap(event: TapEvent): Boolean {
      val image = event.targetElement?.content as? Content.ImageElement ?: return false
      showImagePreview(image); return true
  }
  ```
- New `ImagePreviewDialogFragment` (or Compose dialog): full-screen, loads the image via
  `publication.get(image.embeddedLink)` (reuse existing resource loading / Coil if present),
  shows `image.caption ?: image.text` as caption, dismiss on tap/back. Match the simple Swift
  demo sheet.

## Phase 1 — Docs & changelog
- `CHANGELOG.md`: first **uncomment** the `<!-- ## [Unreleased] -->` heading (line 7), then add
  under Added/Navigator: the new experimental `TapEvent.targetElement`, the linked-image
  limitation, credit the original contributor (grighakobian), and link the guide.
- New guide `docs/guides/navigator/epub-image-preview.md` (Kotlin equivalent of the Swift
  guide): detecting image taps via `InputListener.onTap`, downcasting `targetElement.content`,
  `ImageElement` vs `SvgElement`, and the linked-image limitation. Link from the navigator
  guide index/README. Follow existing guide style (no `---` rules).

## Phase 2 — Compose `web` navigator (follow-up, `readium/navigators/web`)

**Constraint discovered:** `ReflowableWebPublication` (and the FXL equivalent) is **decoupled**
from the shared `Publication` — it exposes only `readingOrder` (href, mediaType), `baseUrl`,
and `container`, with no `linkWithHref`. So the web navigator's `embeddedLink` will be a
**synthesized** `Link(href, mediaType)` built from the relativized `src` + reading-order item,
not a manifest lookup. Confirm exact reachability before starting this phase.

### TypeScript — `internals/scripts/src/`
- In `bridge/all-listener-bridge.ts` `onTap` adapters (`ReflowableListenerAdapter`,
  `FixedListenerAdapter`) and a new `common/` helper, extract the image metadata from
  `event.target`. **Frame coordinates must follow the web navigator's own convention**: CSS px
  → dp (matching `GesturesApi.onTap` building `DpOffset(x.dp, y.dp)` with no pixel-ratio
  scaling), and apply the *same* visual-viewport offset/scale transform already used for the
  tap point (all-listener-bridge.ts lines 44–45). Do **not** reuse the legacy `toNativeRect`
  (×devicePixelRatio) here — that would be wrong by the pixel-ratio factor.
- Reuse only the DOM-walking/caption/selector logic conceptually (separate codebase/language
  from legacy JS). Import `getCssSelector` (dependency present).
- Rebuild: `cd internals/scripts && pnpm install && pnpm run bundle`.

### Native — Kotlin
- `internals/.../webapi/GesturesApi.kt`: parse the metadata in `@JavascriptInterface onTap`
  into a serializable holder (DpRect frame via existing JSON rect helpers); extend
  `GesturesListener.onTap(offset, targetElement?)`.
- `common/.../input/InputListener.kt`: add to the common `TapEvent`:
  `@ExperimentalReadiumApi val targetElement: TargetElement? = null` and
  `@ExperimentalReadiumApi data class TargetElement(val frame: DpRect, val content: Content.Element)`.
- `reflowable/.../resource/ReflowableResource.kt` and `fixedlayout/.../spread/SpreadWebView.kt`
  (`DelegatingGesturesListener` closures): resolve the resource href from the rendered reading
  order item / spread page, synthesize the `Content.Element` (synthesized `Link`), apply the
  same padding shift used for the tap offset.

## Build / regeneration
- Legacy JS: `readium/navigator/src/main/assets/_scripts` → `pnpm install && pnpm run bundle`.
- Web TS (Phase 2): `readium/navigators/web/internals/scripts` → `pnpm install && pnpm run bundle`.
- Kotlin: `./gradlew :readium:shared:assembleDebug :readium:navigator:assembleDebug
  :test-app:assembleDebug` (Phase 1), plus the web modules in Phase 2. Run `ktlintFormat`.

## Verification
1. Build + JS bundle regenerates with no churn beyond the new code.
2. Test app on a reflowable EPUB with **bare** (non-linked) images: tap an `` → full-screen
   preview with correct image + caption; non-image text still toggles system UI; links/footnotes
   still work; text selection unaffected.
3. Caption fallbacks: `alt`, `title`, `<figcaption>`, SVG `<title>`/`<desc>`; `accessibilityLabel`
   from `aria-label`.
4. Inline `<svg>` tap → `Content.SvgElement` with raw SVG; `<svg src=...>`/`` →
   `Content.ImageElement` with resolved `embeddedLink`.
5. Confirm a **linked** image (`<a>`) does **not** fire (documented limitation), and
   `targetElement` is `null` for non-image taps.
5a. Regression check the **decoration** path (highlights/underlines): activating a decoration
    still works after the `clickEvent` JSON shape change (`targetElement` string → object).
6. Confirm `targetElement` requires `@OptIn(ExperimentalReadiumApi::class)` at call sites
   (the test-app listener is the proof).
7. Confirm a legacy **fixed-layout** EPUB still behaves as before (no regression); image
   preview there is expected only via the web navigator in Phase 2.
8. Phase 2: repeat on the web navigator demo (reflowable + FXL).

## Notes
- **Binary compatibility:** adding `targetElement` to `TapEvent` is source-compatible (default
  `null`) but changes the constructor/`copy()`/`componentN` signatures — binary-incompatible for
  published artifacts. Acceptable given the field is `@ExperimentalReadiumApi`; flagged here as a
  conscious choice.
- The legacy `Listener.onTap` signature change is internal (`R2BasicWebView.Listener`); the only
  external API change is the new optional, experimental `TapEvent.targetElement`.
- Images are not in `findNearestInteractiveElement`'s tag list, so bare images already reach
  `onTap` — no change to the interactive-element filter (and that filter is exactly what keeps
  linked images out of scope).

</details>


readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
```diff
@@ -735,6 +738,73 @@ public class EpubNavigatorFragment internal constructor(
...
+        // type, etc.). For resources not in the manifest (e.g. external images), we synthesise a
+        // plain Link.
+        val embeddedLink = (data.src?.let { Url(it) } as? AbsoluteUrl)
+            ?.let { url -> viewModel.internalLinkFromUrl(url) ?: Link(href = url) }
+
+        if (embeddedLink != null) {
+            return Content.ImageElement(
+                locator = locator,
+                embeddedLink = embeddedLink,
+                caption = data.caption,
```
TechWilk: Do these want to be `caption = data.caption?.takeIf { it.isNotBlank() },` to match the accessibilityLabel above?

This one and the SVG one a few lines down.

TechWilk (commented):
I've been through this and it looks sensible. Here are a few small notes but it does work as it is in some prototyping I've been doing (noting the documented limitations)

readium/navigator/src/main/java/org/readium/r2/navigator/epub/EpubNavigatorFragment.kt
mickael-menu: Yes thank you, I'll fix it.

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
+# EPUB Navigators feature parity
+
+This document tracks the feature differences between the legacy EPUB navigator (`EpubNavigatorFragment` in `readium-navigator`) and the new Compose-based web navigators (`ReflowableWebRendition` and `FixedWebRendition` in `readium-navigator-web-reflowable` and `readium-navigator-web-fixedlayout`).
```
mickael-menu, line 3: @qnga I added this document to track differences between the old and new navigators, and to see what's still missing.

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Navigation
+
+| Feature | Legacy | New | Notes |
+|---|:---:|:---:|---|
+| Restoring the initial location | ✅ | ✅ | |
+| Go to locator / link | ✅ | ✅ | New API: `goTo(GoLocation)`, `goTo(Location)`, `goTo(Url)` |
+| Move forward / backward | ✅ | ✅ | `OverflowController.moveForward()` / `moveBackward()`, plus `moveLeft()` / `moveRight()` helpers |
+| `canMoveForward` / `canMoveBackward` | ➖ | ⚠️ | Present but not observable, so Compose UIs may not recompose reliably |
+| Edge-tap page turns | ✅ | ✅ | `DirectionalNavigationAdapter` vs `defaultInputListener()` |
```
qnga, line 31: Animated API page turns have only been added to the old navigator.

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
...
+| Keyboard page turns | ✅ | ❌ | No key event handling at all in the new navigators |
+| Notification when jumping to a distant location | ✅ | ❌ | The legacy `Listener.onJumpToLocator` lets apps record the previous location to offer a "go back" button; no equivalent in the new API |
```
qnga: As discussed once, there is no need for this in the new API. The app can record the jump when it approves it in the hyperlink listener.

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
...
+| Continuous scrolling across resources | ❌ | ⚠️ | Reflowable only, the fixed-layout navigator is always paginated |
+
+## Preferences and settings
+
+| Feature | Legacy | New | Notes |
+|---|:---:|:---:|---|
+| Configurable preferences | ✅ | ✅ | `PreferencesController` in the new API |
+| Preferences editor for building a user settings UI | ✅ | ❌ | No `PreferencesEditor` for the new preference sets |
```
qnga: I would say "Changed" as it's a choice, not a missing feature.

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Reflowable preferences
+
+Present in both: `backgroundColor`, `columnCount`, `fontFamily`, `fontSize`, `fontWeight`, `hyphens`, `imageFilter`, `language`, `letterSpacing`, `ligatures`, `lineHeight`, `paragraphIndent`, `paragraphSpacing`, `readingProgression`, `scroll`, `textAlign`, `textColor`, `textNormalization`, `verticalText`, `wordSpacing`.
+
+| Preference | Legacy | New | Notes |
+|---|:---:|:---:|---|
+| `theme` (light / dark / sepia) | ✅ | 🔄 | Replaced by color preference bundles (`LightTheme`, `SepiaTheme`, `DarkTheme` companion constants) instead of a single enum |
+| `publisherStyles` | ✅ | 🔄 | Replaced by the narrower `overridePublisherColors` |
+| `typeScale` | ✅ | ❌ | |
```
qnga: Deprecated in Readium CSS2. That is not something yet to come.

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
...
+| `pageMargins` | ✅ | 🔄 | Replaced by `minMargins` plus line-length preferences |
+| `linkColor` / `visitedColor` | ❌ | ✅ | New in the Compose navigator |
+| `minimalLineLength` / `optimalLineLength` / `maximalLineLength` | ❌ | ✅ | New adaptive column layout based on line length and system font scale |
+
+### Readium CSS and typography
+
+| Feature | Legacy | New | Notes |
+|---|:---:|:---:|---|
+| Readium CSS injection with RTL / CJK variants | ✅ | ✅ | |
+| Custom `RsProperties` overrides | ✅ | ❌ | Exposed via `Configuration.readiumCssRsProperties` in legacy; `internal` in the new navigator |
```
qnga: In my opinion and as discussed, not something to do.

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
...
+| Custom font family declarations | ✅ | ✅ | Same builder DSL (`fontFamilyDeclarations`) |
+| Bundled accessibility fonts (OpenDyslexic, AccessibleDfA, iA Writer Duospace) | ✅ | ✅ | |
+| `useReadiumCssFontSize` escape hatch | ✅ | ❌ | Legacy only (`DelicateReadiumApi`) |
```
qnga: Not applicable as the "Readium CSS Font Size" no longer exists in Readium CSS 2, as far as I know.

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Decorations
+
+| Feature | Legacy | New | Notes |
+|---|:---:|:---:|---|
+| Apply decoration groups | ✅ | ✅ | Legacy `applyDecorations(list, group)`; new observable `decorations` map on `DecorationController` |
+| Highlight and underline styles | ✅ | ✅ | |
+| Custom styles and HTML templates | ✅ | ✅ | `HtmlDecorationTemplate` vs `WebDecorationTemplate`, same layout / width options |
+| Activable decorations (tap events) | ✅ | ⚠️ | `DecorationListener.onDecorationActivated` exists, but `defaultDecorationListener()` is a no-op placeholder |
```
qnga: ???
Handling the decorations is up to the app, not the toolkit. What would be there if not a no-op placeholder?

docs/guides/navigator/epub-feature-parity.md
```diff
@@ -0,0 +1,132 @@
...
+| Footnote content extraction (`FootnoteContext`) | ✅ | ✅ | |
+| Non-linear resource links | ⚠️ | ⚠️ | New API has a dedicated `onNonLinearLinkActivated` callback but no built-in way to render non-linear resources |
+
+## Locators, progression and positions
+
+| Feature | Legacy | New | Notes |
+|---|:---:|:---:|---|
+| Current location tracking | ✅ | ✅ | `currentLocator: StateFlow` vs observable `location` on the controller |
+| Position and total progression | ✅ | ✅ | FXL positions derived from the reading-order index in the new navigator |
+| `firstVisibleElementLocator()` | ✅ | ❌ | Needed for TTS synchronization |
```
qnga: To start the TTS at the right place, correct? @stevenzeck Did you find an alternative way in your tests?

readium/navigator/src/main/java/org/readium/r2/navigator/input/TapEvent.kt
```diff
@@ -1,12 +1,32 @@
 package org.readium.r2.navigator.input
 
 import android.graphics.PointF
+import android.graphics.RectF
+import org.readium.r2.shared.ExperimentalReadiumApi
+import org.readium.r2.shared.publication.services.content.Content
```
qnga, line 6: I don't think it makes much sense to depend on the classes of a service as it's not related, but we can use this until we've got a better idea of all the places we need something like this.

readium/navigator/src/main/java/org/readium/r2/navigator/input/TapEvent.kt
```diff
@@ -1,12 +1,32 @@
...
+
+/**
+ * A content element targeted by a pointer event, paired with its on-screen [frame].
+ *
+ * @param frame Frame of the element relative to the navigator's view, in device pixels.
+ * @param content The content element under the pointer.
+ */
+@ExperimentalReadiumApi
+public data class TargetElement(
+    val frame: RectF,
```
qnga: The name `frame` doesn't ring a bell. Haven't we been using`rect` so far?
https://github.com/readium/kotlin-toolkit/blob/f8e6f93db81570c7cc0833279b2628f4c65d8efe/readium/navigator/src/main/java/org/readium/r2/navigator/SelectableNavigator.kt#L33

README.md
```diff
@@ -73,6 +73,7 @@ Guides are available to help you make the most of the toolkit.
 * [Navigator](docs/guides/navigator/navigator.md) - an overview of the Navigator to render a `Publication`'s content to the user
 * [Configuring the Navigator](docs/guides/navigator/preferences.md) – setup and render Navigator user preferences (font size, colors, etc.)
 * [Font families in the EPUB navigator](docs/guides/navigator/epub-fonts.md) – support custom font families with reflowable EPUB publications
+* [EPUB image preview](docs/guides/navigator/epub-image-preview.md) – detect when a user taps an image to present it in a dedicated view
```
qnga: Is this feature really about **pre**viewing images? We see first the image in context, and then in a dedicated view.

docs/guides/navigator/epub-feature-parity.md
stevenzeck: For the legacy navigator? Yeah, that's the way I'm aware of. There isn't an equivalent for the new navigator, I just used `NavigationController.location` for now, and if the text is not visible, I go to it via `NavigationController.location`.

Is that what you're asking or am I way off?

qnga: I was asking about your tests with the new navigators. Does `NavigationController.location` contain a location type compatible with the TTS navigator?

stevenzeck: Yeah, it extends `ExportableLocation` so you can do [`NavigationController.location.toLocator`](https://github.com/readium/kotlin-toolkit/blob/d3a8d6f525e3b06976d1b32d6b72b3ec0d3e20a3/readium/navigators/common/src/main/java/org/readium/navigator/common/Locations.kt#L28). `TtsNavigatorFactory.createNavigator` accepts that.

qnga: I think the TTS navigator is able to use only the href from the output of `NavigationController.location.toLocator`. That's why we need `firstVisibleLocator`. It was introduced as an optimisation to avoid computing a text anchor in any location. I wonder if it's really worth it though.

stevenzeck: Yeah, I didn't quite understand where this was going at first. I would say `firstVisibleElementLocator` is more needed for knowing where to start TTS (synchronization isn't the word I would use for that).

Without the cssSelector (which isn't available in the new navigator), TTS has no way to start where we want it to start.

docs/guides/navigator/epub-feature-parity.md
mickael-menu: Done

docs/guides/navigator/epub-feature-parity.md
mickael-menu: Sure

docs/guides/navigator/epub-feature-parity.md
mickael-menu: Sure

docs/guides/navigator/epub-feature-parity.md
mickael-menu: Done

docs/guides/navigator/epub-feature-parity.md
mickael-menu: Right

docs/guides/navigator/epub-feature-parity.md
mickael-menu: There is still a font size in Readium CSS 2: https://github.com/readium/css/blob/develop/docs/CSS12-user_prefs.md#font-size

I don't know yet if we should allow using the Readium CSS font size with the new Compose navigators. Here's the rationale for adding this in the first place: https://github.com/readium/kotlin-toolkit/pull/323

docs/guides/navigator/epub-feature-parity.md
mickael-menu: It's just a misunderstanding.

docs/guides/navigator/epub-feature-parity.md
mickael-menu: I don't know if we need *this* exact API. But we need a way to start the TTS from the currently visible page. I suggest we keep this in the document for now.

mickael-menu: Renamed to "Needed for starting TTS on the current page"

readium/navigator/src/main/java/org/readium/r2/navigator/input/TapEvent.kt
mickael-menu: I also had some doubts about implementing this in the Swift toolkit. But if we don't use `ContentElement`, we need to duplicate this structure for input events and potentially other APIs. It's not that we actually depend on the `Content` service, but I think the `ContentElement` models should be part of `shared.publication.content` and shared across multiple services/APIs.

I have postponed this decision for now, as it will likely be revamped when Guided Navigation support is added.

readium/navigator/src/main/java/org/readium/r2/navigator/input/TapEvent.kt
mickael-menu: Yes it's pulled from the Swift version.

README.md
mickael-menu: I renamed it to "EPUB image viewer"

readium/navigator/src/main/assets/_scripts/src/gestures.js
```diff
@@ -104,3 +106,104 @@ function bindDragGesture(element) {
...
+    imageElement.getAttribute("src") ||
+    imageElement.getAttribute("href") ||
+    null;
+
+  // Resolve the raw src/href attribute to an absolute URL using the document's
+  // base URI. `getAttribute` returns the literal attribute value (possibly
+  // relative), while we need the absolute form so the native side can
+  // relativize it against the publication base URL to recover the correct
+  // manifest href.
+  let src = rawSrc ? new URL(rawSrc, document.baseURI).href : null;
```
stevenzeck, line 150: Should this be in a try/catch in case rawSrc is in an invalid format?

test-app/src/main/java/org/readium/r2/testapp/reader/ImageViewerDialogFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+import org.readium.r2.testapp.utils.compose.AppTheme
+
+/**
+ * A simple full-screen dialog displaying a tapped image and its caption.
+ *
+ * This demonstrates the experimental [org.readium.r2.navigator.input.TapEvent.targetElement] API.
+ */
+@OptIn(ExperimentalReadiumApi::class)
+class ImageViewerDialogFragment(
+    private val image: Content.ImageElement,
```
stevenzeck: Having this as a required parameter will cause exceptions when the fragment is recreated after configuration changes.

test-app/src/main/java/org/readium/r2/testapp/reader/ImageViewerDialogFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+        super.onCreateDialog(savedInstanceState).apply {
+            setCanceledOnTouchOutside(true)
+        }
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?,
+    ): View =
+        ComposeView(requireContext()).apply {
```
stevenzeck, line 78: Probably better to set the `ViewCompositionStrategy` to `DisposeOnViewTreeLifecycleDestroyed`.

test-app/src/main/java/org/readium/r2/testapp/reader/ImageViewerDialogFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+    override fun onStart() {
+        super.onStart()
+        // Make the dialog full-screen.
+        dialog?.window?.setLayout(
+            ViewGroup.LayoutParams.MATCH_PARENT,
+            ViewGroup.LayoutParams.MATCH_PARENT
+        )
+    }
+
+    private suspend fun loadBitmap(): android.graphics.Bitmap? =
```
stevenzeck: If this returns null, ImagePreviewDialogFragment will spin indefinitely.

test-app/src/main/java/org/readium/r2/testapp/reader/ImageViewerDialogFragment.kt
mickael-menu: Done

test-app/src/main/java/org/readium/r2/testapp/reader/ImageViewerDialogFragment.kt
mickael-menu: Fixed

test-app/src/main/java/org/readium/r2/testapp/reader/ImageViewerDialogFragment.kt
mickael-menu: Fixed

readium/navigator/src/main/assets/_scripts/src/gestures.js
mickael-menu: Fixed

stevenzeck (approved):

docs/guides/navigator/epub-feature-parity.md
qnga: Let's make that clearer:
- Readium CSS 1 had a font size based on the font-size CSS property. 
- A different font size strategy based on a native API was added to the Kotlin toolkit.
- Readium CSS 2 has replaced the font size property implementation by a new one based on zoom CSS.
- That's what the new navigator is using underneath the font size preference.
- It doesn't offer the native API-based alternative.

mickael-menu: Ha in that case it's different. Since we're using Readium CSS for handling the font size, this property has no purpose. I removed it from the document.

qnga:
I didn't look at the implementation details, but Steven did. I think we can merge.

PR #805, opened 2026-06-24
Refactor rendition preferences in new navigators

qnga:
- Take preferences instead of settings in rendition controllers
- Remove preferences editors from the toolkit


readium/navigators/web/fixedlayout/src/main/kotlin/org/readium/navigator/web/fixedlayout/FixedWebRenditionState.kt
```diff
@@ -212,15 +216,26 @@ internal data class FixedWebPreloadedData(
...
+    private val settingsResolver =
+        FixedWebSettingsResolver(
+            metadata = publication.metadata,
+            defaults = defaults
+        )
+
+    override var preferences: FixedWebPreferences by mutableStateOf(initialPreferences)
+
+    override val settings: FixedWebSettings
+        get() = settingsResolver.settings(preferences)
```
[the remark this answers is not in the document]
mickael-menu: That was going to be one of my comments 😅 @qnga

demos/navigator/src/main/java/org/readium/demo/navigator/reader/Rendition.kt
```diff
@@ -68,26 +71,30 @@ import org.readium.r2.shared.util.toUri
...
     val showPreferences = remember { mutableStateOf(false) }
     val preferencesSheetState = rememberModalBottomSheetState()
 
     if (showPreferences.value) {
-        ModalBottomSheet(
-            sheetState = preferencesSheetState,
-            onDismissRequest = {
-                showPreferences.value = false
+        readerState.renditionState.controller?.let { controllerNow ->
+            val preferencesViewModel = readerState.createPreferencesViewModel(controllerNow)
```
mickael-menu: It looks like the view model might be recreated on each recomposition. Would it be possible to wrap it in a `remember` call?

demos/navigator/src/main/java/org/readium/demo/navigator/preferences/ReflowablePreferencesViewModel.kt
```diff
@@ -0,0 +1,510 @@
...
+    /**
+     * Enable hyphenation for latin languages.
+     *
+     * Only effective when the layout is LTR.
+     */
+    val hyphens: OptionalBooleanPreference =
+        OptionalBooleanPreferenceDelegate(
+            getValue = { controller.preferences.hyphens },
+            getEffectiveValue = { controller.settings.hyphens },
+            getIsEffective = { true },
```
mickael-menu, line 141: The previous `ReflowableWebPreferencesEditor` calculated `isEffective` for each preference using `ReadiumCssLayout.Stylesheets`. For example, settings like `hyphens`, `letterSpacing`, `wordSpacing`, `ligatures`, `paragraphIndent`, and `textAlign` were only effective for certain stylesheets.

In contrast, the new `ReflowablePreferencesViewModel` hardcodes `getIsEffective = { true }` for these settings, which looks like a regression. Since the controller no longer exposes `ReadiumCssLayout`, the view model currently cannot replicate the original logic. Should we find a way to restore this behavior? Maybe we could encode these rules directly within the toolkit?

readium/navigators/web/fixedlayout/src/main/kotlin/org/readium/navigator/web/fixedlayout/FixedWebRenditionState.kt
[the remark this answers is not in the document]
qnga: The Compose doc discourages using `derivedStateOf` when it's not strictly necessary because it's expensive. I don't master the trade-offs well enough to be sure about what the best strategy is.

readium/navigators/web/reflowable/src/main/kotlin/org/readium/navigator/web/reflowable/ReflowableWebRenditionState.kt
```diff
@@ -142,8 +145,10 @@ public class ReflowableWebRenditionState internal constructor(
 
     internal val layoutDelegate: ReflowableLayoutDelegate =
         ReflowableLayoutDelegate(
-            fontFamilyDeclarations,
-            initialSettings
+            fontFamilyDeclarations = fontFamilyDeclarations,
+            publication = publication,
+            initialPreferences = initialPreferences,
+            defaults = ReflowableWebDefaults()
         )
```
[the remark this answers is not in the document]
qnga: Good point, I missed that.

demos/navigator/src/main/java/org/readium/demo/navigator/reader/Rendition.kt
qnga: Yeah, I thought I did it.

demos/navigator/src/main/java/org/readium/demo/navigator/preferences/ReflowablePreferencesViewModel.kt
qnga: Yeah, I faced this issue: `ReadiumCssLayout` is internal, so it's difficult to know which one is currently in use. My feeling is that no reel app will use this information though, don't you think?

qnga: I'm talking about the settings being ineffective due to the current language. I believe showing that some other settings are disabled, depending on the navigation mode for instance, still makes sense.

readium/navigators/web/fixedlayout/src/main/kotlin/org/readium/navigator/web/fixedlayout/FixedWebRenditionState.kt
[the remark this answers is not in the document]
qnga: Let's do it anyway.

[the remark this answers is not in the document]
mickael-menu: Good question... Do we expect the `settings` value to be read in tight scenarios, such as every tick while scrolling? If so `derivedStateOf` might be helpful despite the warning. Otherwise the current approach should be fine.

demos/navigator/src/main/java/org/readium/demo/navigator/preferences/ReflowablePreferencesViewModel.kt
qnga: If someone is really interested, they could check the language setting.

mickael-menu: I don't think we necessarily need to fix this in this PR, but we should keep it in mind if we ever encode the rules in the toolkit again.

I think we should at least add a documentation comment for all impacted preferences. This will ensure users know what to expect if these preferences do not work with specific publications.

Using the language directly might be a bit complicated though. If I remember correctly, the rules determining when they are enabled are complex.

qnga: Ideally, I think these rules should be applied in every stylesheet, for the sake of consistency and in case of multiple languages. We'll see later, with Jiminy.

readium/navigators/web/fixedlayout/src/main/kotlin/org/readium/navigator/web/fixedlayout/FixedWebRenditionState.kt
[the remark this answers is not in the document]
qnga: The settings were computed on every composition, I made the change.

demos/navigator/src/main/java/org/readium/demo/navigator/preferences/ReflowablePreferencesViewModel.kt
qnga: Regarding the doc, I'm on the fence right now. Preferences are a so complex topic that a guide might be better suited than doc comments.

mickael-menu: Agreed. We could write a proper guide and mention it in the doc comments so people don't miss it.

mickael-menu (approved):
Looking good, thank you for making the changes

PR #808, opened 2026-07-08
Fix poorly-authored EPUB links not navigating

mickael-menu:
#### Shared

* EPUB HREFs that are not percent-encoded but carry a fragment or query (e.g. `chapter one.xhtml#section`, with a space in the filename) now keep their `#fragment`/`?query` instead of encoding the separators into the path. This fixes table of contents and Media Overlays links failing to resolve and navigate in poorly-authored EPUBs.

---

Here's a sample EPUB that can be used to reproduce the issue: 
[unencoded-hrefs.epub.zip](https://github.com/user-attachments/files/29794898/unencoded-hrefs.epub.zip)

Ported from https://github.com/readium/swift-toolkit/pull/847

### DIFF
--- a/readium/shared/src/main/java/org/readium/r2/shared/publication/Link.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/publication/Link.kt
@@ -207,16 +207,27 @@ public data class Link(
 }
 
 /**
- * Returns the first [Link] with the given [href], or null if not found.
+ * Returns the index of the first [Link] with the given [href], or null if not found.
+ *
+ * Falls back to ignoring the query and fragment of [href], to stay consistent with
+ * [Manifest.linkWithHref].
  */
-public fun List<Link>.indexOfFirstWithHref(href: Url): Int? =
-    indexOfFirst { it.url().isEquivalent(href) }
-        .takeUnless { it == -1 }
+public fun List<Link>.indexOfFirstWithHref(href: Url): Int? {
+    fun index(href: Url): Int? =
+        indexOfFirst { it.url().isEquivalent(href) }.takeUnless { it == -1 }
+
+    return index(href)
+        ?: index(href.removeQuery().removeFragment())
+}
 
 /**
- * Finds the first link matching the given HREF.
+ * Finds the first [Link] with the given [href], or null if not found.
+ *
+ * Falls back to ignoring the query and fragment of [href], to stay consistent with
+ * [Manifest.linkWithHref].
  */
-public fun List<Link>.firstWithHref(href: Url): Link? = firstOrNull { it.url().isEquivalent(href) }
+public fun List<Link>.firstWithHref(href: Url): Link? =
+    indexOfFirstWithHref(href)?.let { this[it] }
 
 /**
  * Finds the first link with the given relation.
--- a/readium/shared/src/main/java/org/readium/r2/shared/publication/Link.kt
+++ b/readium/shared/src/main/java/org/readium/r2/shared/publication/Link.kt
@@ -209,7 +209,7 @@ public data class Link(
 /**
  * Returns the index of the first [Link] with the given [href], or null if not found.
  *
- * Falls back to ignoring the query and fragment of [href], to stay consistent with
+ * Falls back to ignoring the fragment, then the query too, of [href], to stay consistent with
  * [Manifest.linkWithHref].
  */
 public fun List<Link>.indexOfFirstWithHref(href: Url): Int? {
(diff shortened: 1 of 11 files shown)


readium/shared/src/main/java/org/readium/r2/shared/publication/Link.kt
```diff
@@ -207,16 +207,27 @@ public data class Link(
...
  */
-public fun List<Link>.indexOfFirstWithHref(href: Url): Int? =
-    indexOfFirst { it.url().isEquivalent(href) }
-        .takeUnless { it == -1 }
+public fun List<Link>.indexOfFirstWithHref(href: Url): Int? {
+    fun index(href: Url): Int? =
+        indexOfFirst { it.url().isEquivalent(href) }.takeUnless { it == -1 }
+
+    return index(href)
+        ?: index(href.removeQuery().removeFragment())
```
qnga: As discussed a few weeks ago, we should also look for the href with the query and without the fragment.

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
```diff
@@ -388,10 +388,32 @@ public fun Url.Companion.fromLegacyHref(href: String): Url? =
...
+    // is left in the suffix rather than mistaken for a query separator.
+    val splitIndex = href.indexOfFirst { it == '?' || it == '#' }
+    val pathPart = if (splitIndex < 0) href else href.substring(0, splitIndex)
+    val suffix = if (splitIndex < 0) "" else href.substring(splitIndex)
+
+    val pathUrl = fromDecodedPath(pathPart) ?: return null
+
+    // The suffix is reattached verbatim. Note this can degrade for malformed input: `Url()` parses
+    // via Java's stricter `URI`, so a raw space inside the query/fragment (e.g. `foo.xhtml#a b`)
+    // fails to re-parse and we fall back to `pathUrl` without the suffix.
```
qnga: If the path is not correctly percent-encoded, the query and fragment are likely not to be encoded either. We could encode them as well, instead of ignoring them, don't you think?

readium/shared/src/main/java/org/readium/r2/shared/publication/Link.kt
mickael-menu: Ha right, I'll fix that on Friday.

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
mickael-menu: I wondered about this, but I'm not so sure. A fragment or query is always inherently percent-encoded because of the parsing rules we discussed a while ago. If we percent-encode it again, we risk double-encoding and breaking a valid fragment. In practice, an HTML ID is unlikely to have special characters anyway.

Without actual data, these are only assumptions unfortunately.

qnga: What about the same policy we use for the path? Try to make a valid URL of it, if it fails, encode it.

mickael-menu: Yes I guess we could do that

readium/shared/src/main/java/org/readium/r2/shared/util/Url.kt
```diff
@@ -388,10 +389,51 @@ public fun Url.Companion.fromLegacyHref(href: String): Url? =
...
+
+    // Otherwise the HREF is not a valid URL (e.g. a decoded path containing spaces). Split off the
+    // query and fragment, so each component can be encoded independently. The path ends at the first
+    // `?` or `#`: since the query always precedes the fragment in a URL, a `?` occurring inside a
+    // fragment is left in the suffix rather than mistaken for a query separator.
+    val splitIndex = href.indexOfFirst { it == '?' || it == '#' }
+    val pathPart = if (splitIndex < 0) href else href.substring(0, splitIndex)
+    val suffix = if (splitIndex < 0) "" else href.substring(splitIndex)
+
+    val pathUrl = fromDecodedPath(pathPart) ?: return null
```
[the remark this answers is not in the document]
qnga: This sounds good, doesn't it?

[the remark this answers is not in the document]
mickael-menu: I've addressed the other comments, but this one feels like an over-correction. Do we really expect to encounter HREFs where the path is percent-encoded but the fragments are not (`dir/chapter%20one.xhtml#foo bar`)? I also wonder if we actually see unencoded fragments in the first place.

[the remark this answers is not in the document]
qnga: It's unlikely for fragments, but for queries? What's the point of not handling this case? Performance?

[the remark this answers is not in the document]
mickael-menu: Same for queries. I think this is unnecessary. I'd rather keep the code simple and focused.

Now if this is blocking for you, I can add this bit.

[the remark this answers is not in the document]
qnga: It seems simpler to me to follow the same policy for each component, as claimed in the doc comment, but it might be personal. As you like.

qnga (approved):

Issue #811, opened 2026-07-11
Pdfium adapter: currentLocator is one page ahead of the visible page (page-index off-by-one in PdfiumDocumentFragment)


huttarl:
### Describe the bug

With the pdfium adapter, PdfNavigatorFragment.currentLocator reports a position one page ahead of the page actually on screen: viewing page 1 of a 6-page PDF yields locations.position == 2, page 2 yields 3, and so on.

Expectation: currentLocator.locations.position (1-based) should match the visible page: on page 1, position = 1.

Cause (detailed in Additional context): PdfiumDocumentFragment's convertPageIndexToView / convertPageIndexFromView treat the navigator-facing page index as 1-based, but the contract everywhere else — including elsewhere in the same file — is 0-based. So the +/-1 conversion shifts every reported page by one.

Consequences:

1. Every location emission is one page high. Reading-progress tracking built on currentLocator credits the wrong page throughout.
2. The first page is never reported. Its position appears only as the StateFlow's transient initial value, replaced ~tens of ms after layout, without user interaction.
3. The last page produces no locator update at all. The fragment reports pageIndex == pageCount, and PdfNavigatorViewModel.onPageChanged does positions().getOrNull(pageCount), but it's out of range so it's silently dropped, so currentLocator still shows the second-to-last page.
4. go(locator) lands one page early. The navigator converts the 1-based position to a 0-based index, then convertPageIndexToView subtracts 1 again.

Why it's easy to miss: saving currentLocator and restoring it via initialLocator round-trips to the correct page, because the +1 on save and the −1 on restore cancel. The error only becomes visible when the locator is consumed in absolute terms (progress/analytics, position display, jumping to a locator from elsewhere).

### How to reproduce?

Reproduced with the stock PdfNavigatorFragment + PdfiumEngineProvider (no customizations) hosted in our app; the same applies to the Test App since the code involved is entirely in the toolkit:

1. Open any multi-page PDF with the pdfium adapter.
2. Observe currentLocator, e.g.:
navigator.currentLocator.collect { locator ->
    Log.d("repro", "position=${locator.locations.position}")
}
3. While page 1 is visible, the last emission is position=2 (after a transient position=1 that fires and is replaced within tens of milliseconds, with no user interaction).
4. Swipe forward one page at a time: visible page N always reports position=N+1.
5. On the last page, no emission occurs at all (see consequence 3 above).

Actual log from a 6-page PDF (positions vs. the page actually on screen):

```
[...]

PR #812, opened 2026-07-11
Fixed page positions off by one in the PDFium adapter

huttarl:
The navigator-facing pageIndex contract has been 0-based since the PDF engine extraction in #129: PdfNavigatorFragment previously subtracted 1 before setting the locator value, and when that logic moved to PdfNavigatorViewModel the subtraction was dropped. The conversion methods moved into PdfiumDocumentFragment in the same refactoring kept the old 1-based assumption, so currentLocator reported every page one too high, the first page was never reported, and the last page's locator update was dropped.

Rename the converters' parameters to make the two 0-based conventions explicit (navigator pageIndex vs. PDFView viewPageIndex). The PSPDFKit adapter is unaffected (it passes PSPDFKit's already-0-based index through without conversion).

Fixes issue #811.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,17 +22,13 @@ All notable changes to this project will be documented in this file. Take a look
 
 #### Navigator
 
-* Fixed the Pdfium navigator reporting page positions off by one: `currentLocator` was
-  one page ahead of the visible page, the first page was never reported, and the last
-  page never updated `currentLocator` ([#812](https://github.com/readium/kotlin-toolkit/pull/812)).
-    * **You must migrate persisted `Locator` objects created by the Pdfium navigator**
-      (bookmarks, reading progression) — stored positions were one page too high. Use the
-      new `Publication.migrateLegacyPdfiumLocator()` helper in `readium-adapter-pdfium`,
-      and take a look at [the migration guide](docs/migration-guide.md).
-      
+* Fixed the PDFium adapter reporting page positions off by one: `currentLocator` was one page ahead of the visible page, the first page was never reported, and the last page never updated `currentLocator` (contributed by [@huttarl](https://github.com/readium/kotlin-toolkit/pull/812)).
+    * :warning: You must migrate the `Locator` objects created by the PDFium adapter and persisted in your database (e.g. bookmarks, reading progression), as their positions were one page too high. Use the new `Publication.migrateLegacyPdfiumLocator()` helper and take a look at [the migration guide](docs/migration-guide.md).
+
 #### Shared
 
 * EPUB HREFs that are not percent-encoded but carry a fragment or query (e.g. `chapter one.xhtml#section`, with a space in the filename) now keep their `#fragment`/`?query` instead of encoding the separators into the path. This fixes table of contents and Media Overlays links failing to resolve and navigate in poorly-authored EPUBs.
+
 
 ## [3.3.0] - 2026-06-02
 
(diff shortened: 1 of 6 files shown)


mickael-menu (commented):
Thank you for reporting this issue. I can confirm that I am able to reproduce it and that your PR successfully addresses the fix.

I have some reservations about merging this change in its current form, however. Modifying this now would break bookmarks and app progression for users who have relied on the Pdfium adapter since 2022.

The PDFium adapter is now considered legacy and is unofficially deprecated. We recommend using PSPDFKit, and we plan to integrate the newer Jetpack PDF for our free/open-source adapter eventually. That said, if locators created with PDFium are not fixed, they will break whenever a user decides to switch to a different adapter.

I believe we should move forward with this fix, but we also need to include a migration guide section. Ideally, we should provide helpers to help users convert their Locators during the migration. This PR could be useful as inspiration for those additions: https://github.com/readium/kotlin-toolkit/pull/399

huttarl:
Thanks for looking at the PR and #811. I see your point about breaking bookmarks and app progression for existing users. But you're right that if not fixed, the issue could cause breakage for those who switch to other adapters.

Sorry to hear that the Pdfium adapter is deprecated! I just finished a fairly involved and much-anticipated migration away from PSPDFKit to Pdfium. :-/ I hope the Jetpack PDF integration comes soon! Maybe I can contribute to it.

mickael-menu:
I'm curious why migrating from PSPDFKit to Pdfium was so involved for your app? It should be a mostly drop-in replacement when switching adapters, aside from the preferences, which differ for each.

If you're interested in contributing an adapter for Jetpack PDF, that would be very welcome!

huttarl:
Why migrating from PSPDFKit to Pdfium was so involved for us -- because of non-code-related issues, like researching the available options, finding a suitable one, and deciding whether the benefits were worth the costs (like not having annotations).

I will see about contributing an adapter for Jetpack PDF. That would be beneficial for us, and I would be glad to give back to Readium. We're also using the Readium EPUB components.

huttarl:
Update to my last comment: given that Jetpack PDF is still in alpha with minSdk 28, it
seems premature to develop an adapter now — the API is still changing, and we'd risk a
good bit of rewrite. Open to contributing one when Jetpack PDF reaches beta.

From a quick look at the current Jetpack PDF API: the high-level PdfViewerFragment doesn't seem to expose
enough control for an adapter (no page observation/navigation; a password can only be supplied via its
built-in dialog). But the newer low-level PdfView plus SandboxedPdfLoader (which accepts
a password programmatically) look sufficient.

- Separately: I'm working now on a migration guide for this PR.

huttarl:
I added a migration guide for this change to docs/migration-guide.md, and also updated CHANGELOG.

You mentioned providing helpers. I included helper code in the migration guide, but do you want me to put it in a real API (e.g. an extension in the pdfium adapter module)? Similar to Url.fromLegacyHref() in #399.

stevenzeck:
> I will see about contributing an adapter for Jetpack PDF. That would be beneficial for us, and I would be glad to give back to Readium. We're also using the Readium EPUB components.

I started a [proof of concept](https://github.com/stevenzeck/kotlin-toolkit/tree/androidx-pdf-view) adapter for Jetpack several months ago. As you noted, it's still in alpha, and I don't like the current solution for streaming PDFs. See this issue in Google's issue tracker: https://issuetracker.google.com/issues/358918885.

huttarl:
Following up: I've added the conversion helper for stored Locators as an extension function called `Publication.migrateLegacyPdfiumLocator()`, in the pdfium navigator module, following the pattern of `Url.fromLegacyHref()` from #399. The migration guide and CHANGELOG now point to this function instead of an inline snippet. If you'd rather keep the guide-snippet-only approach, I'm happy to drop this API. Otherwise I believe this covers everything from your review. Let me know if there's anything else you need from me.

mickael-menu (approved):
Thank you @huttarl!

